PR #47, opened 2018-03-21
Replaces manually defined default constructors in bsfUtility/Utility …

jonesmz:
…with the default keyword, and some misc other changes.

Feel free to object to any of these changes. I'm happy to rework the PR.

### DIFF
--- a/Source/Foundation/bsfUtility/Utility/BsBitwise.h
+++ b/Source/Foundation/bsfUtility/Utility/BsBitwise.h
@@ -63,7 +63,8 @@ namespace bs
 	{
 	public:
 		/** Returns the most significant bit set in a value. */
-		static UINT32 mostSignificantBitSet(unsigned int value)
+		template<typename IntType>
+		static constexpr UINT32 mostSignificantBitSet(IntType value)
 		{
 			UINT32 result = 0;
 			while (value != 0) {
--- a/Source/Foundation/bsfUtility/Utility/BsOctree.h
+++ b/Source/Foundation/bsfUtility/Utility/BsOctree.h
@@ -18,9 +18,7 @@ namespace bs
 	class OctreeElementId
 	{
 	public:
-		OctreeElementId()
-			:node(nullptr), elementIdx(0)
-		{ }
+		OctreeElementId() = default;
 
 		OctreeElementId(void* node, UINT32 elementIdx)
 			:node(node), elementIdx(elementIdx)
@@ -30,8 +28,8 @@ namespace bs
 		template<class, class>
 		friend class Octree;
 
-		void* node;
-		UINT32 elementIdx;
+		void* node = nullptr;
+		UINT32 elementIdx = 0u;
 	};
 
 	/** 
@@ -114,10 +112,8 @@ namespace bs
 			{ }
 
 			HChildNode(UINT32 index)
-				:index(index)
-			{
-				empty = false;
-			}
+				:index(index), empty(false)
+			{ }
 		};
 
 		/** Contains a range of child nodes in an octree node. */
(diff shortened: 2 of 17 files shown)


Source/Foundation/bsfUtility/Utility/BsUtil.cpp
```diff
@@ -8,7 +8,7 @@ namespace bs
 	String md5(const WString& source)
 	{
 		MD5 md5;
-		md5.update((UINT8*)source.c_str(), (UINT32)source.length() * sizeof(WString::value_type));
+		md5.update((UINT8*)source.data(), (UINT32)source.length() * sizeof(WString::value_type));
```
jonesmz: It's theoretically allowed for c_str() to involve an allocation and copy. Since md5.update() is being given the length as well as the data, just calling data() is perfectly safe, and avoids the possibility of an STL implementation allocating under the hood.

Source/Foundation/bsfUtility/Utility/BsBitwise.h
```diff
@@ -63,7 +63,8 @@ namespace bs
 	{
 	public:
 		/** Returns the most significant bit set in a value. */
-		static UINT32 mostSignificantBitSet(unsigned int value)
+		template<typename IntType>
```
jonesmz: This avoids the possibility of weird integer promotion / narrowing problems.

BearishSun:
Also looking good! Nice to see things getting cleaned up.

Source/Foundation/bsfUtility/Utility/BsOctree.h
```diff
@@ -331,7 +325,7 @@ namespace bs
 			/** Calculates bounds for the provided child node. */
 			NodeBounds getChild(HChildNode child) const
 			{
-				static constexpr float map[] = { -1.0f, 1.0f };
+				static constexpr const float map[2] = { -1.0f, 1.0f };
```
BearishSun: Is there an advantage of using const here? Also the specific statement of array size seems redundant perhaps.

Source/Foundation/bsfUtility/Utility/BsOctree.h
```diff
@@ -210,10 +203,11 @@ namespace bs
 			NodeElements mElements;
 
 			Node* mParent;
-			Node* mChildren[8];
+			Node* mChildren[8] = {  nullptr, nullptr, nullptr, nullptr,
+									nullptr, nullptr, nullptr, nullptr };
 
-			UINT32 mTotalNumElements : 31;
-			UINT32 mIsLeaf : 1;
+			UINT32 mTotalNumElements : 31 = 0;
```
BearishSun: This bit didn't compile on newest MSVC, not sure what the standard says about that.

Source/Foundation/bsfUtility/Utility/BsOctree.h
jonesmz: We can drop this changed line, that's no problem.

I'm a fan of specificity when it comes to array sizes, but given it's only 2 items I understand why most people wouldn't care to bother.

As for the constexpr vs const, I'm not 100% certain that there is or isn't a difference. My understanding of the standard is that constexpr variable are still modifiable in some situations in compile-time contexts, but to be honest I haven't put in the due-diligence to actually verify that one way or another, and I don't know of any examples where that could happen.

Feel free to drop this change.

Source/Foundation/bsfUtility/Utility/BsOctree.h
jonesmz: Hrmm, weird. I was able to use this on gcc no problem.

Perhaps it's a gcc extension? 

I'll keep that in mind for later changes.

PR #49, opened 2018-03-21
Represent [U]INT* with the types from <cstdint>. Resolves #48.

### LINKED ISSUE #48: Use <cstdint>in bsfUtility/BsTypes.h
Currently BsTypes.h's typedefs are defined as such.

    namespace bs
    {
        typedef char INT8;
        typedef unsigned char UINT8;
        typedef short INT16;
        typedef unsigned short UINT16;
        typedef int INT32;
        typedef unsigned int UINT32;

    #if BS_COMPILER == BS_COMPILER_MSVC
        typedef unsigned __int64 UINT64;
        typedef __int64 INT64;
    #else
        typedef unsigned long long UINT64;
        typedef long long INT64;
    #endif
    }

We can replace these with the types in <cstdint>

    namespace bs
    {
        typedef std::int8_t INT8;
        typedef std::uint8_t UINT8;
        typedef std::int16_t INT16;
        typedef std::uint16_t UINT16;
        typedef std::int32_t INT32;
        typedef std::uint32_t UINT32;
        typedef std::int64_t UINT64;
        typedef std::uint64_t INT64;
    }

jonesmz:


Source/Foundation/bsfUtility/String/BsString.cpp
```diff
@@ -199,29 +199,6 @@ namespace bs
 		return stream.str();
 	}
 
-	WString toWString(long val,
```
jonesmz: Need to remove these because they collide with the UINT64 versions.

PR #51, opened 2018-03-21
Refactoring to use default keyword, and other minor changes, for Source/Foundation

jonesmz:
Feel free to ask for changes.

### DIFF
--- a/Source/Foundation/bsfUtility/Reflection/BsIReflectable.cpp
+++ b/Source/Foundation/bsfUtility/Reflection/BsIReflectable.cpp
@@ -35,10 +35,9 @@ namespace bs
 	RTTITypeBase* IReflectable::_getRTTIfromTypeId(UINT32 rttiTypeId)
 	{
 		Stack<RTTITypeBase*> todo;
-		Vector<RTTITypeBase*>& rootClasses = getDerivedClasses();
 
-		for(auto iter = rootClasses.begin(); iter != rootClasses.end(); ++iter)
-			todo.push(*iter);
+		for(const auto & item : getDerivedClasses())
+			todo.push(item);
 
 		while(!todo.empty())
 		{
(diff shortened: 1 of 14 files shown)


Source/Foundation/bsfUtility/Reflection/BsIReflectable.cpp
```diff
@@ -35,10 +35,9 @@ namespace bs
 	RTTITypeBase* IReflectable::_getRTTIfromTypeId(UINT32 rttiTypeId)
 	{
 		Stack<RTTITypeBase*> todo;
-		Vector<RTTITypeBase*>& rootClasses = getDerivedClasses();
 
-		for(auto iter = rootClasses.begin(); iter != rootClasses.end(); ++iter)
-			todo.push(*iter);
+		for(const auto & item : getDerivedClasses())
```
BearishSun: I prefer to keep the & next to the type, like so `const auto&`. I'm in the process of merging right now so I'll modify it on my end.

Source/Foundation/bsfUtility/Serialization/BsSerializedObject.h
```diff
@@ -38,12 +38,10 @@ namespace bs
...
 	{
-		SerializedEntry()
-			:fieldId(0), serialized(nullptr)
-		{ }
+		SerializedEntry() = default;
 
-		UINT32 fieldId;
-		SPtr<SerializedInstance> serialized;
+		UINT32 fieldId = 0;
+		SPtr<SerializedInstance> serialized{nullptr};
```
BearishSun: This one is redundant I believe, shared pointers are null by default.

jonesmz:
Thanks for the info on the reference token. I'll try to keep that in mind.

Have you considered using something like clang-format, so automate formatting decisions like this?

Sure, I think you're right on the nullptr. I was kind of doing most of these changes mechanically, not really strongly thinking, while some other stuff was chugging on another computer. I probably did some other silly things as well.

BearishSun:
That's a good suggestion, I added a new issue for it https://github.com/GameFoundry/bsf/issues/55.

PR #54, opened 2018-03-21
Refactoring to use default keyword, and other minor changes, for Source/Foundation/Math

jonesmz:
Feel free to request changes.

### DIFF
--- a/Source/Foundation/bsfUtility/Math/BsQuaternion.h
+++ b/Source/Foundation/bsfUtility/Math/BsQuaternion.h
@@ -22,16 +22,13 @@ namespace bs
 		};
 
 	public:
-		Quaternion()
-		{ }
+		Quaternion() = default;
+		Quaternion(const Quaternion&) = default;
+		Quaternion& operator=(const Quaternion&) = default;
 
-		Quaternion(BS_ZERO zero)
-			:x(0.0f), y(0.0f), z(0.0f), w(0.0f)
-		{ }
+		Quaternion(BS_ZERO zero) { }
 
-		Quaternion(BS_IDENTITY identity)
-			:x(0.0f), y(0.0f), z(0.0f), w(1.0f)
-		{ }
+		Quaternion(BS_IDENTITY) :w(1.0f) { }
 
 		Quaternion(float w, float x, float y, float z)
 			:x(x), y(y), z(z), w(w)
@@ -185,15 +182,6 @@ namespace bs
 		/** Gets the positive z-axis of the coordinate system transformed by this quaternion. */
 		Vector3 zAxis() const;
 
-		Quaternion& operator= (const Quaternion& rhs)
-		{
-			w = rhs.w;
-			x = rhs.x;
-			y = rhs.y;
-			z = rhs.z;
-
-			return *this;
-		}
 
 		Quaternion operator+ (const Quaternion& rhs) const 
 		{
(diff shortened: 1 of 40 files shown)


Source/Foundation/bsfUtility/Math/BsQuaternion.h
```diff
@@ -360,15 +348,19 @@ namespace bs
...
 		static Quaternion getRotationFromTo(const Vector3& from, const Vector3& dest, const Vector3& fallbackAxis = Vector3::ZERO);
 
-		static const float EPSILON;
+		static constexpr const float EPSILON = 1e-03f;
 
 		static const Quaternion ZERO;
 		static const Quaternion IDENTITY;
 
-		float x, y, z, w; // Note: Order is relevant, don't break it
+		float x = 0.0f;
```
BearishSun: For Quaternion, and a few other classes, like Vector, we don't want default value initialization. They can be used in very performance critical code where these values might get changed right away, so it ends up being extra instructions. These classes generally have specialized constructors that accept BS_ZERO or similar, that allow the user to zero-initialize if wanted.

jonesmz:
The compiler should only be initializing them once, based on which constructor is chosen.

I think the standard indicates that this initialization to zero is happening whether or not we explicitly provide a default value of zero.

I believe that floats are always "value-initialized" unless provided a value for "brace-or-equal-initializer", so this these floats are always initialized to 0.0f by default. 

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf

Page 221

```
8 To value-initialize an object of type T means:
(8.1) — if T is a (possibly cv-qualified) class type (Clause 12) with either no default constructor (15.1) or a default constructor that is user-provided or deleted, then the object is default-initialized;
(8.2) — if T is a (possibly cv-qualified) class type without a user-provided or deleted default constructor, then the object is zero-initialized and the semantic constraints for default-initialization are checked, and if T has a non-trivial default constructor, the object is default-initialized;
(8.3) — if T is an array type, then each element is value-initialized;
(8.4) — otherwise, the object is zero-initialized
```



Have you had a chance to compare the assembly of the Quaternion classes on https://godbolt.org/ ?

jonesmz:
Some more discussion here: https://stackoverflow.com/questions/3803153/what-are-primitive-types-default-initialized-to-in-c

BearishSun:
I'm not sure if we understand each other. I meant, no initialization at all (leave them at their undefined values). This way the compiler can completely skip calling the constructor and initializing. Same if I just `malloc`d the memory.

BearishSun:
You can check what I mean here: https://godbolt.org/g/RTM8J9 . Try removing the in-class initialization to zero in the Quaternion class, for all four members.

jonesmz:
If you change to O2, the constructor is eliminated as well: https://godbolt.org/g/Y5pGic

I have to admit that I don't see how the constructor isn't being called to value-initialize the floats. How is that permitted? Do you know why clang isn't emitting a call to the default constructor?

jonesmz:
Bizarrely, adding brace-initilization changes the assembly output. Wtf? https://godbolt.org/g/ZYPo3J

Though, that's a bit off-topic.

jonesmz:
If the source of setValues() is available (e.g. eligible for inlining, not in another translation unit), at O2, the compiler will directly initialize to the values from setValues()

https://godbolt.org/g/KTzeBd

BearishSun:
By default floats/ints are not value-initialized. They are not initialized at all. So the constructor does nothing unless you explictly default initialize, or initialize them to some value.

Regarding brace init, I'm guessing this is happening:
> If the number of initializer clauses is less than the number of members or initializer list is completely empty, the remaining members are value-initialized. If a member of a reference type is one of these remaining members, the program is ill-formed.

But I'm hardly well versed in the standard.

BearishSun:
It's not just going to be a problem in -O1. Just using an array is enough to foil the compiler, even at -O2:
https://godbolt.org/g/NzMXRd

jonesmz:
In that example, GCCs producing some pretty disgusting assembly...

Clang handles that scenario beautifully though.

https://godbolt.org/g/eiEZdy

jonesmz:
And https://godbolt.org/g/1wubJT, for when we provide default values.

jonesmz:
Here's with gcc : https://godbolt.org/g/fTXypA

jonesmz:
Increasing the array size to 1024, where the only other difference is that one is initialized to 0, the other isn't given a value, has vastly different behavior for gcc vs clang.

https://godbolt.org/g/oCKZif

Gcc inserts a call to memset for the code that initializes the floats to zero, but clang's assembly actually shrinks significantly. Looks like it's just adding up all the values and returning them statically.

BearishSun:
I'm too tired to read all this assembly :) But indeed, some really weird results.

I think my main concern with initialization for these types were cases where there are many of those objects being allocated, thousands or more, in some low-level high performance code, and then passed along to something the compiler might not be able to inline.

jonesmz:
To be honest here, my biggest concern was that we wouldn't be able to construct the Quaternion in constexpr contexts.

I should have started the discussion with that.

However, apparently I was misremembering the rules, and it's perfectly fine to do this (as just an example):

struct Quaternion
{
    float x;// = 0.0f;
    float y;// = 0.0f;
    float z;// = 0.0f;
    float w;// = 0.0f;
    Quaternion() = default;
    constexpr Quaternion(float)
     : x(0.0f)
     , y(0.0f)
     , z(0.0f)
     , w(0.0f)
     {}
};

So, best of both worlds.

"no-initialization" for arrays and such.
Constexpr zero-initialization when asked for.

Everyone wins.

Lets leave it how you have it.

PR #64, opened 2018-03-31
Small refactor BsThreadPool.

jonesmz:
Apparently my branch name is wrong. Misread the diff when I created the branch. Oh well.

### DIFF
--- a/Source/Foundation/bsfUtility/Threading/BsTaskScheduler.cpp
+++ b/Source/Foundation/bsfUtility/Threading/BsTaskScheduler.cpp
@@ -5,27 +5,26 @@
 
 namespace bs
 {
-	Task::Task(const PrivatelyConstruct& dummy, const String& name, std::function<void()> taskWorker, 
+	Task::Task(const PrivatelyConstruct& dummy, String name, std::function<void()> taskWorker,
 		TaskPriority priority, SPtr<Task> dependency)
-		: mName(name), mPriority(priority), mTaskId(0), mTaskWorker(taskWorker), mTaskDependency(dependency), mState(0)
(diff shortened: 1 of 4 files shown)


Source/Foundation/bsfUtility/Threading/BsTaskScheduler.cpp
```diff
@@ -5,27 +5,26 @@
...
+		: mName(std::move(name)), mPriority(priority), mTaskWorker(std::move(taskWorker)), mTaskDependency(std::move(dependency))
 	{
 
 	}
 
-	SPtr<Task> Task::create(const String& name, std::function<void()> taskWorker, TaskPriority priority, SPtr<Task> dependency)
+	SPtr<Task> Task::create(String name, std::function<void()> taskWorker, TaskPriority priority, SPtr<Task> dependency)
 	{
-		return bs_shared_ptr_new<Task>(PrivatelyConstruct(), name, taskWorker, priority, dependency);
+		return bs_shared_ptr_new<Task>(PrivatelyConstruct(), std::move(name), std::move(taskWorker), priority, std::move(dependency));
```
BearishSun: I believe we need a separate constructor accepting the rvalues returned by std::move. In this case I'd prefer we just stick with normal references rather than creating more constructors. In fact that's why I have been avoiding passing ownership by value + std::move in most cases, it just creates a daisy chain of extra code and it's easy to mess up by forgetting a move constructor and suddenly your code is doing copies when it should be just passing addresses around.

BearishSun:
Thank you, merged. Made changes regarding value-passing and an `const auto&` iterator, let me know if you disagree about those.

PR #66, opened 2018-03-31
Generalizes smart pointer template functions, and some other minor refactoring

jonesmz:
This PR is needed for my DynLib PR

PR #67, opened 2018-03-31
Refactor dynlib to use std::set for pointer storage

jonesmz:
Requires https://github.com/GameFoundry/bsf/pull/66/

### DIFF
--- a/Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
+++ b/Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
@@ -5,27 +5,29 @@
 
 namespace bs
 {
-	DynLib* DynLibManager::load(const String& name)
+	DynLib* DynLibManager::load(String filename)
 	{
 		// Add the extension (.dll, .so, ...) if necessary.
-		String filename = name;
-		const UINT32 length = (UINT32)filename.length();
+		// TODO: The string comparison here could be slightly more efficent
+		// Using a templatized string_concat function for the lower_bound call
+		// and/or a custom comparitor that does comparison by parts.
+		const String::size_type length = filename.length();
 		const String extension = String(".") + DynLib::EXTENSION;
-		const UINT32 extLength = (UINT32)extension.length();
+		const String::size_type extLength = extension.length();
 		if (length <= extLength || filename.substr(length - extLength) != extension)
-			filename += extension;
+			filename.append(extension);
 
 		if(DynLib::PREFIX != nullptr)
-			filename = DynLib::PREFIX + filename;
+			filename.insert(0, DynLib::PREFIX);
 
-		auto iterFind = mLoadedLibraries.find(filename);
+		const auto& iterFind = mLoadedLibraries.find(filename);
 		if (iterFind != mLoadedLibraries.end())
 		{
 			return iterFind->second;
 		}
 		else
 		{
-			DynLib* newLib = new (bs_alloc<DynLib>()) DynLib(filename);
+			DynLib* newLib = new (bs_alloc<DynLib>()) DynLib(std::move(filename));
 			mLoadedLibraries[filename] = newLib;
 
 			return newLib;
@@ -34,7 +36,7 @@ namespace bs
 
 	void DynLibManager::unload(DynLib* lib)
 	{
-		auto iterFind = mLoadedLibraries.find(lib->getName());
+		const auto& iterFind = mLoadedLibraries.find(lib->getName());
 		if (iterFind != mLoadedLibraries.end())
 		{
 			mLoadedLibraries.erase(iterFind);
--- a/Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
+++ b/Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
@@ -20,16 +20,15 @@ namespace bs
 		if(DynLib::PREFIX != nullptr)
 			filename.insert(0, DynLib::PREFIX);
 
-		const auto& iterFind = mLoadedLibraries.find(filename);
-		if (iterFind != mLoadedLibraries.end())
+		const auto& iterFind = mLoadedLibraries.lower_bound(filename);
+		if (iterFind != mLoadedLibraries.end() && iterFind->first == filename)
 		{
 			return iterFind->second;
 		}
 		else
 		{
-			DynLib* newLib = new (bs_alloc<DynLib>()) DynLib(std::move(filename));
-			mLoadedLibraries[filename] = newLib;
-
+			DynLib* newLib = new (bs_alloc<DynLib>()) DynLib(filename);
+			mLoadedLibraries.emplace_hint(iterFind, std::move(filename), newLib);
 			return newLib;
 		}
 	}
--- a/Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
+++ b/Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
@@ -1,10 +1,24 @@
 //************************************ bs::framework - Copyright 2018 Marko Pintera **************************************//
 //*********** Licensed under the MIT license. See LICENSE.md for full terms. This notice is not to be removed. ***********//
 #include "Utility/BsDynLibManager.h"
-#include "Utility/BsDynLib.h"
 
 namespace bs
 {
+	static inline bool operator<(const NPtr<DynLib>& lhs, const NPtr<DynLib>& rhs)
+	{
+		return lhs->getName() < rhs->getName();
+	}
+
+	static inline bool operator<(const NPtr<DynLib>& lhs, const String & rhs)
+	{
+		return lhs->getName() < rhs;
+	}
+
+	static inline bool operator<(const String & lhs, const NPtr<DynLib>& rhs)
+	{
+		return lhs < rhs->getName();
+	}
+
 	DynLib* DynLibManager::load(String filename)
 	{
 		// Add the extension (.dll, .so, ...) if necessary.
@@ -21,14 +35,14 @@ namespace bs
 			filename.insert(0, DynLib::PREFIX);
 
 		const auto& iterFind = mLoadedLibraries.lower_bound(filename);
-		if (iterFind != mLoadedLibraries.end() && iterFind->first == filename)
+		if (iterFind != mLoadedLibraries.end() && (*iterFind)->getName() == filename)
 		{
-			return iterFind->second;
+			return iterFind->get();
 		}
 		else
 		{
-			DynLib* newLib = new (bs_alloc<DynLib>()) DynLib(filename);
-			mLoadedLibraries.emplace_hint(iterFind, std::move(filename), newLib);
+			DynLib* newLib = new (bs_alloc<DynLib>()) DynLib(std::move(filename));
+			mLoadedLibraries.emplace_hint(iterFind, newLib);
 			return newLib;
 		}
 	}
@@ -50,8 +64,8 @@ namespace bs
 		// Unload & delete resources in turn
 		for(auto& entry : mLoadedLibraries)
 		{
-			entry.second->unload();
-			bs_delete(entry.second);
+			entry->unload();
+			bs_delete(entry.get());
 		}
 
 		// Empty the list
--- a/Source/Foundation/bsfUtility/Utility/BsDynLibManager.h
+++ b/Source/Foundation/bsfUtility/Utility/BsDynLibManager.h
@@ -5,16 +5,16 @@
 #include "Prerequisites/BsPrerequisitesUtil.h"
 #include "Utility/BsModule.h"
 
-namespace bs 
+namespace bs
 {
 	/** @addtogroup General
 	 *  @{
 	 */
 
 	/**
-	 * This manager keeps a track of all the open dynamic-loading libraries, opens them and returns references to 
+	 * This manager keeps a track of all the open dynamic-loading libraries, opens them and returns references to
 	 * already-open libraries.
-	 *			
+	 *
 	 * @note	Not thread safe.
 	 */
 	class BS_UTILITY_EXPORT DynLibManager : public Module<DynLibManager>
@@ -28,7 +28,7 @@ namespace bs
 		 *
 		 * @param[in]	name	The name of the library. The extension can be omitted.
 		 */
-		DynLib* load(const String& name);
+		DynLib* load(String name);
 
 		/** Unloads the given library. */
 		void unload(DynLib* lib);
--- a/Source/Foundation/bsfUtility/Utility/BsDynLibManager.h
+++ b/Source/Foundation/bsfUtility/Utility/BsDynLibManager.h
@@ -5,6 +5,8 @@
 #include "Prerequisites/BsPrerequisitesUtil.h"
 #include "Utility/BsModule.h"
 
+#include "Utility/BsDynLib.h"
+
 namespace bs
 {
 	/** @addtogroup General
@@ -34,7 +36,7 @@ namespace bs
 		void unload(DynLib* lib);
 
 	protected:
-		Map<const String, DynLib*> mLoadedLibraries;
+		Set<NPtr<DynLib>, std::less<>> mLoadedLibraries;
 	};
 
 	/** Easy way of accessing DynLibManager. */
--- a/Source/Foundation/bsfUtility/Utility/BsDynLibManager.h
+++ b/Source/Foundation/bsfUtility/Utility/BsDynLibManager.h
@@ -34,7 +34,7 @@ namespace bs
 		void unload(DynLib* lib);
 
 	protected:
-		Map<String, DynLib*> mLoadedLibraries;
+		Map<const String, DynLib*> mLoadedLibraries;
 	};
 
 	/** Easy way of accessing DynLibManager. */
--- a/Source/Foundation/bsfUtility/Prerequisites/BsStdHeaders.h
+++ b/Source/Foundation/bsfUtility/Prerequisites/BsStdHeaders.h
@@ -195,9 +195,9 @@ namespace bs
 	}
 
 	/**
-	* Create a new unique pointer from a previously constructed object.
-	* Pointer specific data will be allocated using the provided allocator category.
-	*/
+	 * Create a new unique pointer from a previously constructed object.
+	 * Pointer specific data will be allocated using the provided allocator category.
+	 */
 	template<typename Type, typename Alloc = GenAlloc, typename Deleter = decltype(&bs_delete<Type, Alloc>)>
 	UPtr<Type, Alloc, Deleter> bs_unique_ptr(Type* data, Deleter del = &bs_delete<Type, Alloc>)
 	{
@@ -212,6 +212,25 @@ namespace bs
 
 		return bs_unique_ptr<Type, Alloc, Deleter>(rawPtr);
 	}
+
+	template<typename T>
+	struct NativePtr
+	{
+		constexpr NativePtr(T* p) : m_p(p) {}
+		constexpr T& operator*(void) const { return *m_p; }
+		constexpr T* operator->(void) const { return m_p; }
+		constexpr T* get(void) const { return m_p; }
+	private:
+		T* m_p = nullptr;
+	};
+
+	/**
+	 * "Smart" pointer that is not smart. Does nothing but hold a pointer value. No memory management is performed at all.
+	 * This class exists to make storing pointers in containers easier to manage, such with non-member comparison operators.
+	 */
+	template <typename T>
+	using NPtr = NativePtr<T>;
+
 
 	/** @} */
 }
--- a/Source/Foundation/bsfUtility/Utility/BsDynLib.cpp
+++ b/Source/Foundation/bsfUtility/Utility/BsDynLib.cpp
@@ -17,11 +17,9 @@
 
 namespace bs
 {
-	DynLib::DynLib(const String& name)
+	DynLib::DynLib(String name)
+		:mName(std::move(name))
 	{
-		mName = name;
-		mHandle = nullptr;
-
 		load();
 	}
 
@@ -76,14 +74,15 @@ namespace bs
 			0,
 			NULL
 		);
-		String ret = (char*)lpMsgBuf;
+
+		String ret((char*)lpMsgBuf);
 		// Free the buffer.
 		LocalFree(lpMsgBuf);
 		return ret;
 #elif BS_PLATFORM == BS_PLATFORM_LINUX || BS_PLATFORM == BS_PLATFORM_OSX
 		return String(dlerror());
 #else
-		return String("");
+		return String();
 #endif
 	}
 }
--- a/Source/Foundation/bsfUtility/Utility/BsDynLib.h
+++ b/Source/Foundation/bsfUtility/Utility/BsDynLib.h
@@ -9,7 +9,7 @@
 	typedef struct HINSTANCE__* hInstance;
 #endif
 
-namespace bs 
+namespace bs
 {
 	/** @addtogroup General
 	 *  @{
@@ -30,7 +30,7 @@ namespace bs
 #endif
 
 	/** Class that holds data about a dynamic library. */
-	class BS_UTILITY_EXPORT DynLib
+	class BS_UTILITY_EXPORT DynLib final
 	{
 	public:
 		/** Platform-specific file extension for a dynamic library (e.g. "dll"). */
@@ -56,7 +56,7 @@ namespace bs
 #endif
 
 		/** Constructs the dynamic library object and loads the library with the specified name. */
-		DynLib(const String& name);
+		DynLib(String name);
 		~DynLib() = default;
 
 		/** Loads the library. Does nothing if library is already loaded. */
@@ -72,7 +72,7 @@ namespace bs
 		 * Returns the address of the given symbol from the loaded library.
 		 *
 		 * @param[in] strName	The name of the symbol to search for.
-		 * @return				If the function succeeds, the returned value is a handle to the symbol. Otherwise null.
+		 * @return		If the function succeeds, the returned value is a handle to the symbol. Otherwise null.
 		 */
 		void* getSymbol(const String& strName) const;
 
@@ -83,8 +83,8 @@ namespace bs
 		String dynlibError();
 
 	protected:
-		String mName;
-		DYNLIB_HANDLE mHandle;
+		const String mName;
+		DYNLIB_HANDLE mHandle = nullptr;
 	};
 
 	/** @} */


Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
```diff
@@ -1,60 +1,44 @@
...
-			filename += extension;
+			filename.append(extension);
 
 		if(DynLib::PREFIX != nullptr)
-			filename = DynLib::PREFIX + filename;
+			filename.insert(0, DynLib::PREFIX);
 
-		auto iterFind = mLoadedLibraries.find(filename);
-		if (iterFind != mLoadedLibraries.end())
+		const auto& iterFind = mLoadedLibraries.lower_bound(filename);
```
BearishSun: I don't believe this should be a reference, considering the iterator is returned by value, unless I'm misunderstanding which type this gets deduced to.

Source/Foundation/bsfUtility/Utility/BsDynLibManager.h
```diff
@@ -5,37 +5,70 @@
...
+			void operator()(DynLib* lib) const
+			{
+				lib->unload();
+				bs_delete<DynLib, GenAlloc>(lib);
+			}
+		};
+
+		using DynLibPtr_t = UPtr<DynLib, GenAlloc, DynLibDeleter>;
+
+		struct DynLibPtrComparator final
```
BearishSun: I'd prefer we keep the Map as it was. It's a lot of extra code, and in the header file as well - to me it just looks a lot uglier without much upsides.

Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
```diff
@@ -1,60 +1,44 @@
...
 		{
-			return iterFind->second;
+			return iterFind->get();
 		}
 		else
 		{
-			DynLib* newLib = new (bs_alloc<DynLib>()) DynLib(filename);
-			mLoadedLibraries[filename] = newLib;
-
+			DynLib* newLib = new (bs_alloc<DynLib>()) DynLib(std::move(filename));
```
BearishSun: We probably want a move constructor for DynLib to avoid a copy here.

BearishSun:
This doesn't compile on MSVC (#66 changes were included). Perhaps it's the allocator changes? I remember having compiler specific issues when it came to their definition.

```
2>c:\projects\bansheeenginedev\bsf\source\foundation\bsfutility\allocators\bsmemoryallocator.h(435): error C2280: 'std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>::unique_ptr(const std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter> &)': attempting to reference a deleted function (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\memory(2394): note: see declaration of 'std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>::unique_ptr' (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xmemory0(811): note: see reference to function template instantiation 'void bs::StdAlloc<U,Alloc>::construct<_Ty,_Ty&>(std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter> *,_Ty &)' being compiled
2>        with
2>        [
2>            U=std::_Tree_node<std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,void *>,
2>            Alloc=bs::GenAlloc,
2>            _Ty=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xmemory0(811): note: see reference to function template instantiation 'void bs::StdAlloc<U,Alloc>::construct<_Ty,_Ty&>(std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter> *,_Ty &)' being compiled
2>        with
2>        [
2>            U=std::_Tree_node<std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,void *>,
2>            Alloc=bs::GenAlloc,
2>            _Ty=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xmemory0(819): note: see reference to function template instantiation 'void std::_Normal_allocator_traits<_Alloc>::_Construct1<_Ty,_Ty&>(std::false_type,_Alloc &,_Ty *,_Ty &)' being compiled
2>        with
2>        [
2>            _Alloc=bs::StdAlloc<std::_Tree_node<std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,void *>,bs::GenAlloc>,
2>            _Ty=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xmemory0(818): note: see reference to function template instantiation 'void std::_Normal_allocator_traits<_Alloc>::_Construct1<_Ty,_Ty&>(std::false_type,_Alloc &,_Ty *,_Ty &)' being compiled
2>        with
2>        [
2>            _Alloc=bs::StdAlloc<std::_Tree_node<std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,void *>,bs::GenAlloc>,
2>            _Ty=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xtree(774): note: see reference to function template instantiation 'void std::_Normal_allocator_traits<_Alloc>::construct<_Ty,_Ty&>(_Alloc &,_Ty *,_Ty &)' being compiled
2>        with
2>        [
2>            _Alloc=bs::StdAlloc<std::_Tree_node<std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,void *>,bs::GenAlloc>,
2>            _Ty=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xtree(773): note: see reference to function template instantiation 'void std::_Normal_allocator_traits<_Alloc>::construct<_Ty,_Ty&>(_Alloc &,_Ty *,_Ty &)' being compiled
2>        with
2>        [
2>            _Alloc=bs::StdAlloc<std::_Tree_node<std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,void *>,bs::GenAlloc>,
2>            _Ty=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xtree(1906): note: see reference to function template instantiation 'std::_Tree_node<_Ty,void *> *std::_Tree_comp_alloc<_Traits>::_Buynode<_Ty&>(_Ty &)' being compiled
2>        with
2>        [
2>            _Ty=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,
2>            _Traits=std::_Tset_traits<bs::DynLibManager::DynLibPtr_t,bs::DynLibManager::DynLibPtrComparator,bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>,false>
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xtree(1906): note: see reference to function template instantiation 'std::_Tree_node<_Ty,void *> *std::_Tree_comp_alloc<_Traits>::_Buynode<_Ty&>(_Ty &)' being compiled
2>        with
2>        [
2>            _Ty=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,
2>            _Traits=std::_Tset_traits<bs::DynLibManager::DynLibPtr_t,bs::DynLibManager::DynLibPtrComparator,bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>,false>
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xtree(1932): note: see reference to function template instantiation 'std::_Tree_node<_Ty,void *> *std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::_Copy_or_move<_Value_type,std::integral_constant<bool,true>>(_Ty &,std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::_Copy_tag,_Is_set)' being compiled
2>        with
2>        [
2>            _Ty=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,
2>            _Kty=bs::DynLibManager::DynLibPtr_t,
2>            _Pr=bs::DynLibManager::DynLibPtrComparator,
2>            _Alloc=bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>,
2>            _Value_type=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,
2>            _Is_set=std::integral_constant<bool,true>
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xtree(1932): note: see reference to function template instantiation 'std::_Tree_node<_Ty,void *> *std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::_Copy_or_move<_Value_type,std::integral_constant<bool,true>>(_Ty &,std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::_Copy_tag,_Is_set)' being compiled
2>        with
2>        [
2>            _Ty=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,
2>            _Kty=bs::DynLibManager::DynLibPtr_t,
2>            _Pr=bs::DynLibManager::DynLibPtrComparator,
2>            _Alloc=bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>,
2>            _Value_type=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,
2>            _Is_set=std::integral_constant<bool,true>
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xtree(1888): note: see reference to function template instantiation 'std::_Tree_node<_Ty,void *> *std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::_Copy_nodes<_Moveit>(std::_Tree_node<_Ty,void *> *,std::_Tree_node<_Ty,void *> *,_Moveit)' being compiled
2>        with
2>        [
2>            _Ty=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,
2>            _Kty=bs::DynLibManager::DynLibPtr_t,
2>            _Pr=bs::DynLibManager::DynLibPtrComparator,
2>            _Alloc=bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>,
2>            _Moveit=std::_Tree<std::_Tset_traits<bs::DynLibManager::DynLibPtr_t,bs::DynLibManager::DynLibPtrComparator,bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>,false>>::_Copy_tag
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xtree(1888): note: see reference to function template instantiation 'std::_Tree_node<_Ty,void *> *std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::_Copy_nodes<_Moveit>(std::_Tree_node<_Ty,void *> *,std::_Tree_node<_Ty,void *> *,_Moveit)' being compiled
2>        with
2>        [
2>            _Ty=std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>,
2>            _Kty=bs::DynLibManager::DynLibPtr_t,
2>            _Pr=bs::DynLibManager::DynLibPtrComparator,
2>            _Alloc=bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>,
2>            _Moveit=std::_Tree<std::_Tset_traits<bs::DynLibManager::DynLibPtr_t,bs::DynLibManager::DynLibPtrComparator,bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>,false>>::_Copy_tag
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xtree(984): note: see reference to function template instantiation 'void std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::_Copy<std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::_Copy_tag>(const std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>> &,_Moveit)' being compiled
2>        with
2>        [
2>            _Kty=bs::DynLibManager::DynLibPtr_t,
2>            _Pr=bs::DynLibManager::DynLibPtrComparator,
2>            _Alloc=bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>,
2>            _Moveit=std::_Tree<std::_Tset_traits<bs::DynLibManager::DynLibPtr_t,bs::DynLibManager::DynLibPtrComparator,bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>,false>>::_Copy_tag
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xtree(984): note: see reference to function template instantiation 'void std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::_Copy<std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::_Copy_tag>(const std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>> &,_Moveit)' being compiled
2>        with
2>        [
2>            _Kty=bs::DynLibManager::DynLibPtr_t,
2>            _Pr=bs::DynLibManager::DynLibPtrComparator,
2>            _Alloc=bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>,
2>            _Moveit=std::_Tree<std::_Tset_traits<bs::DynLibManager::DynLibPtr_t,bs::DynLibManager::DynLibPtrComparator,bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>,false>>::_Copy_tag
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xtree(978): note: while compiling class template member function 'std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>> &std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::operator =(const std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>> &)'
2>        with
2>        [
2>            _Kty=bs::DynLibManager::DynLibPtr_t,
2>            _Pr=bs::DynLibManager::DynLibPtrComparator,
2>            _Alloc=bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\set(146): note: see reference to function template instantiation 'std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>> &std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::operator =(const std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>> &)' being compiled
2>        with
2>        [
2>            _Kty=bs::DynLibManager::DynLibPtr_t,
2>            _Pr=bs::DynLibManager::DynLibPtrComparator,
2>            _Alloc=bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\set(54): note: see reference to class template instantiation 'std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>' being compiled
2>        with
2>        [
2>            _Kty=bs::DynLibManager::DynLibPtr_t,
2>            _Pr=bs::DynLibManager::DynLibPtrComparator,
2>            _Alloc=bs::StdAlloc<bs::DynLibManager::DynLibPtr_t,bs::GenAlloc>
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\projects\bansheeenginedev\bsf\source\foundation\bsfutility\utility\bsdynlibmanager.h(70): note: see reference to class template instantiation 'std::set<bs::DynLibManager::DynLibPtr_t,bs::DynLibManager::DynLibPtrComparator,bs::StdAlloc<T,bs::GenAlloc>>' being compiled
2>        with
2>        [
2>            T=bs::DynLibManager::DynLibPtr_t
2>        ] (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
2>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\memory(2394): note: 'std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter>::unique_ptr(const std::unique_ptr<bs::DynLib,bs::DynLibManager::DynLibDeleter> &)': function was explicitly deleted (compiling source file C:\Projects\BansheeEngineDev\bsf\Source\Foundation\bsfCore\Audio\BsAudioManager.cpp)
```

jonesmz:
I'll rework the pull request :-)

Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
jonesmz: The reference should decay so that the resulting compiled code is simply storing the value.

The main value of using const-& when using the auto keyword is in it's behavior in the general case of various situations.

When the function returns a value, the const-auto-& does "lifetime extension", so that the temporary that's returned lives until the const-auto-& variable goes out of scope. In effect, after the compiler gets done with it, this really just means that the variable is treated as a value, not a reference, and the "lifetime extension of the temporary" is only meaningful in the land of standarese, not in the land of assembly code.

When the function returns a reference, the const-auto-& captures that reference, and prevents any kind of copy-construction (which can potentially involve an allocation, depending on underlying type, and malloc is expensive and blocking). So the avoidance of the copy-construction is desirable, except in the case where we *want* a copy, so that the original object that the reference points to can be modified.

So perhaps it's a bit of a force of habit on my part, but I almost always prefer to use const auto&, or auto&& when storing the return value of a function, except in cases where I explicitly need a copy, and the reason is specifically because in my normal day-to-day work, the codebase I work with isn't consistent on whether a function returns by value, or by reference, and whether copy-construction is cheap or not. For what I do, it's safer to use the "safe" default of const-auto-& than to try to micro manage what I do with the returned variable.

That being said, yes, in this situation, the & doesn't meaningfully change what the compiler is going to do, as the returned value is, i believe, a value.

Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
jonesmz: I'm not sure I know what you mean.

What's being copied? The pointer? Or something else?

Source/Foundation/bsfUtility/Utility/BsDynLibManager.h
jonesmz: I think there are performance benefits to converting to "Set" instead of "Map".

Two somewhat distinct things.

1) Set allows us to avoid the allocate-and-copy of an additional String. If we had a read-only-reference-counted string, i'd say that it was a meaningless distinction between the two, and wouldn't care. But allocations and copies are expensive in general (though, how expensive is subjective and context sensitive).

2) If we convert to Set, but otherwise don't care about switching to using the "lower_bound" function to save on comparisons when iterating over the Map/Set, then we can use a comparitor that only has the single operator(). However, in order to support the use of the lower_bound function, which I'm almost certain has better performance when needing to insert new values (because it skips a lot of iterating on the insert_hint call), we need to provide a transparent comparitor so that we can search directly for the string, without having allocated a DynLib* already.

I suppose it would be possible to make a rough implementation of lower_bound directly in the cpp file (where lower_bound is called now), and skip the need for the extra two operator() functions. Would you rather do that?

Source/Foundation/bsfUtility/Utility/BsDynLibManager.h
```diff
@@ -5,37 +5,70 @@
...
 		void unload(DynLib* lib);
 
 	protected:
-		Map<String, DynLib*> mLoadedLibraries;
-    };
+		struct DynLibDeleter final
+		{
+			void operator()(DynLib* lib) const
+			{
+				lib->unload();
```
jonesmz: Does using UPtr seem good to you? Or would you also rather not do that?

I wonder if perhaps, instead of using UPtr, we should use SPtr inside of DynLibManager, and std::weak_ptr (WPtr?) for the returned-type ? Not 100% sure if that would preserve the semantics, as I think weak_ptr can be upgraded to strong_ptr. But as long as the allocator used to allocate the DynLib* stays alive longer than the SPtr does, I don't think it'll hurt anything.

Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
BearishSun: I wasn't aware of this const-& behavior and assumed it's just going to end up being a plain reference. This is definitely fine then!

jonesmz: You can read more about it here, if you sift through the discussion: https://stackoverflow.com/questions/19864176/how-would-auto-extend-the-life-time-of-the-temporary-object

Source/Foundation/bsfUtility/Utility/BsDynLibManager.h
BearishSun: I'm aware this was done for performance reasons, but it's highly unlikely DynLib is ever going to be on a performance critical path. When it's not performance critical I value cleaner/smaller code and design in general.

But if you can move the implementation in the .cpp file, in any form, I'd be happy with that.

Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
BearishSun: The `DynLib(std::move(filename))`. Since DynLib doesn't have a constructor accepting a String&&, this will invoke a copy on the filename, unless I'm mistaken.

Source/Foundation/bsfUtility/Utility/BsDynLibManager.h
jonesmz: That's very reasonable.

Let me clarify my behavior / motivation here a little bit. I'm mostly a "Can't see the forest for the trees" kind of person, and I have a difficult time understanding how to use something if I haven't taken it apart and understood the relationships between all the internal parts.

So what I've been doing is going through the foundational code for bsf, file by file, and making sure I understand how it functions piece by piece.

So all of these PRs that I've been sending in aren't intended as any kind of criticism or anything. I'm just trying to understand how everything fits together and I figured I might as well try to feed some (relatively minor) improvements back to you while I do it.

So it's really not a huge deal for me, one way or another, if you don't want to accept a particular PR. The act of creating the PR was itself the goal, so that I could understand the code.

That being said, sure, I'll try to move more of the logic into the cpp file. I agree with your stance on trying to keep header files simple and small.

Source/Foundation/bsfUtility/Utility/BsDynLibManager.h
BearishSun: I don't like UPtrs because of having to create custom deleters. I'd prefer it over SPtr because of the potential upgrade to strong ptr you mention, it just sounds dirty even if it technically couldn't happen :)

In a simple case like this I'd personally still go with `bs_new`/`bs_delete`. But I'm not against UPtr if you think the extra safety is important.

BearishSun: Actually using SPtr for storage, but returning a raw pointer instead of a weak one, might be the best case?

Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
jonesmz: I think you're mistaken.

Part of my day job involves (in addition to network protocols) dealing with my employers super low level framework code. E.g. Custom smart pointers. fundemental data types, that kind of thing. We even have a fork of STLPort (an STL implementation) that we use internally, I'm kind of the accidental maintainer for it (ugh, it's so ugly to work with).

Anyway, the way rvalue-references work, from a human-level perspective, is that they're supposed to signal permission to the function that's receiving the rvalue-reference that data can be "stolen" from the object passed in as an rvalue-reference.

Unless someone's using an absolutely ancient STL implementation (Here's looking at you, STLPort!@#$), their std::string (or rather, std::basic_string) will have a "move-constructor".

The implementation's *probably* going to look something like this (simplified, but you'll understand).

```
std::basic_string::basic_string(basic_string && other)
 : m_size(other.m_size)
 , m_pointer(other.m_pointer)
#ifdef debug
 , m_endpointer(other.m_endpointer)
#endif
{
    other.m_size = 0;
    other.m_pointer = nullptr;
#ifdef debug
    other.m_endpointer = nullptr;
#endif
}
```

Compare that to a copy-constructor, which is going to look something like this


```
std::basic_string::basic_string(const basic_string & other)
{
    m_pointer = new char[other.m_size+1];
    std::copy(other.m_pointer, other.m_pointer+other.m_size, m_pointer);
    m_size = other.m_size;
#ifdef debug
.....
#endif
}
```

So when a function accepts an object by value, instead of by const-& or &&, that function is saying unequivocal "I am going to own an instance of whatever you pass in, either by copy, or by move. It's up to you to decide what that looks like".

It allows the caller of the function to decide the nature of the data the function owns. If the caller wants the function to own the data by copy, it simply passes the value in like normal.

If, however, the callers no longer wants to own the data, it uses std::move() to pass the data in.

jonesmz: There are some semantics when it comes to how references are treated, of course, that also change the decision here slightly.

When dealing with a function that's inline, in the header file, passing by const-& or && really makes no difference. As long as the compiler is "inlining" the function (which it almost certainly will if it's a small function) the reference-ness collapses completely, and you end up with just value semantics at the assembly code level.

But there are two situations where that's not the case.

1) Cross dll boundaries,  sometimes. The compiler generally doesn't want to assume that the implementation of the function stays the same when it's across dll boundaries, or at the very least, you do end up with a version of the function being instantiated in the elf shared library, regardless of inlining or not.

For those instantiations, passing the value by const& **always** requires a pointer  (for those platforms that implement references as pointers, which is basically all platforms).

Of course, that's not too horrible. Generally a pointer  once at the beginning of the function is no big deal. But when you're planning to copy the data (via a call to the copy constructor) anyway, that's a pointer dereference that you didn't strictly speaking NEED to do, if you had let the callee make the copy of the data for you.

Same with functions that are declared in the header file, and then implemented in the cpp file. Since inlining is greatly prevented, you always pay the pointer dereference, regardless. If you're function is planning to OWN the data being passed in, why not just say so to the code that's calling your function?

jonesmz: Of course, all of this is irrelevant for types that both 1) Do not have a move constructor, and also 2) Should have a move constructor.

In that case, accepting the parameter by value will likely be more expensive.

BearishSun: Makes perfect sense, I blanked on the fact that String's move constructor would be called before being passed to DynLib's constructor.

jonesmz: Somewhat related to that, I once raised a question on stack overflow about why variables that the compiler knew were about to die aren't automatically converted to rvalue-references at their last use.

https://stackoverflow.com/questions/45763943/compiler-deduction-of-rvalue-references-for-variables-going-out-of-scope

I think there's some merit in trying to get the C++ Standards committee to consider that for a future version, even though there are (currently) some places where that'll cause problems.

jonesmz:
I've totally rewritten this PR (By rewriting the commit history, so previous comments might not line up).

I think you'll find the first 4 commits in the PR to be just fine.

The 5th commit doesn't resolve your concerns about the Set comparitor type, I haven't figured out the best way to handle moving the comparison operations into the cpp file just yet.

BearishSun:
Yep the changes look good! I'm much happier with this form of UPtr.

jonesmz:
@BearishSun 

You'll probably like the latest version.

I forgot that std::less<> doesn't need the operator< to be globally defined, just available at the point of use.

Now the .h file is as simple as it can possibly get, and the cpp file has all the implementation details.

jonesmz:
Ok, visual studio has a broken implementation of... something.

Going to drop unique ptr from this change set.

BearishSun:
Yep, looks good. I made some very minor code style changes, but otherwise merged as is. Not sure what the error with MSVC is, I checked it quickly but it wasn't obvious. But it's well known for not following the standard as strictly as GCC/Clang, so no surprise I guess.

jonesmz:
From what I could tell, their implementation of std::set *always* uses the copy constructor at compile time, but possibly not runtime.

They choose to manually impement std set's copy-constructor, instead of setting it to = default;. That appears to always generate the code to call the value_types copy-constructor (in this case, std::unique_ptr) which doesn't exist.

Clang / gcc seems to set the std::set constructors and assignment operators to = default, which behaves better by not always generating that code even if it doesn't exist.

I think anyway.

As far as I can tell, there's no way to work around that. std::set and std::unique_ptr simply can't be used together on MSVC. That's *probably* a bug, but reporting it to Microsoft won't do us any good. They'll almost certainly ignore it *shrug*.

PR #73, opened 2018-04-01
Enable -Wextra, fix existing warnings. Resolves #68.

### LINKED ISSUE #68: Consider enabling -Wextra
I enable -Wextra in all of my projects. While the warnings that it generates aren't quite as dangerous to ignore as -Wall, resolving those warnings tends to result in a codebase that's quite a bit more robust to hard to solve problems.

jonesmz:


BearishSun:
Looks good, thanks.

PR #79, opened 2018-04-09
Use the variadic form of StdAlloc::construct for platforms that support it

jonesmz:
Avoids inadvertent copy-construction.


BearishSun:
Looks good, merged as is.

PR #80, opened 2018-04-09
Refactor the integral data types, and some casting, for BsMemoryAllocator.h

jonesmz:
Mostly it's removing casts that I don't see the benefit of, and normalizing the integer data types used as function parameters. E.g. using size_t when counting "number objects" and such.


BearishSun:
Looks good, merged as is.

PR #82, opened 2018-04-09
Minor documentation change to BsDynLibManager.h

jonesmz:


BearishSun:
Thanks, merged. Note that I try to keep documentation text to wrap to a new line at column 125 (same length as the header text in every file).

PR #84, opened 2018-04-10
Initialize RTTI's mInitMembers object with 'this' by default, 

jonesmz:
removing the need to initialize it in every RTTI-able object.

Took significantly less time than I anticipated.

I'd like to point out again that I really enjoy how clean and modern you've kept the bsf code. So many projects out there are... well, unreadable, i guess is the most polite way to say it.

Bsf / Banshee is beautiful by comparison to almost everything else I've seen out there.

Take, as an example, the Ninja build system. I've been using it at work, and wanted to hack some compatibility features into it. It's... it's bad.

I've got a couple of patches to it that clean it up a little bit using C++11 features, but the maintainers have no interest in using anything newer than C++98. So I'm probably going to fork it (Maybe name it Splinter, after the famous ninja master from the popular cartoon involving turtles). *shrug*.

Banshee / Bsf is just so damn pretty by comparison.

### DIFF
--- a/Documentation/Manuals/User/serializingObjects.md
+++ b/Documentation/Manuals/User/serializingObjects.md
@@ -145,15 +145,13 @@ public:
 		BS_RTTI_MEMBER_PLAIN(myString, 2)
 	BS_END_RTTI_MEMBERS
 
-	MyComponentRTTI()
-		:mInitMembers(this)
-	{ }
+	MyComponentRTTI() = default;
 	
 	// ... 
 };
 ~~~~~~~~~~~~~
 
-Field definition portion of the RTTI type always begins with the @ref bs::BS_BEGIN_RTTI_MEMBERS "BS_BEGIN_RTTI_MEMBERS" macro, and ends with the @ref bs::BS_END_RTTI_MEMBERS "BS_END_RTTI_MEMBERS". If field definition is provided you must also provide a constructor that initializes the *mInitMembers* with the *this* pointer.
+Field definition portion of the RTTI type always begins with the @ref bs::BS_BEGIN_RTTI_MEMBERS "BS_BEGIN_RTTI_MEMBERS" macro, and ends with the @ref bs::BS_END_RTTI_MEMBERS "BS_END_RTTI_MEMBERS".
 
 The field members themselves are defined by calling macros starting with BS_RTTI_MEMBER_*. The macro expects the name of the field it describes, as well as a unique ID of the field. The suffix of the BS_RTTI_MEMBER_* macro depends on the type of the field being added. There are three different types:
  - @ref bs::BS_RTTI_MEMBER_PLAIN "BS_RTTI_MEMBER_PLAIN" - Field containing basic data types like ints, floats, strings or other types that can be just trivially copied during serialization/deserialization.
@@ -191,9 +189,7 @@ public:
 		BS_RTTI_MEMBER_REFL(mesh, 6)
 	BS_END_RTTI_MEMBERS
 
-	MyComponentRTTI()
-		:mInitMembers(this)
-	{ }
+	MyComponentRTTI() = default;
 	
 	// ... 
 };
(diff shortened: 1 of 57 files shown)


jonesmz:
Btw, kind of unrelated to this PR, but I noticed when looking at the CI systems that they're downloading the dependencies almost every time (every time?).

I know that at least for Travis, there's some way to pre-upload the dependencies to their system, so the download doesn't keep happening (or at least only happens internally, instead of across the internet).

Not sure if that's something you want to address or not. I imagine bandwidth from github.com isn't a big deal.

BearishSun:
Thanks for the compliments, nice to hear it's being appreciated. And it's getting prettier by the day with your help :)

I think I investigated caching for CI, but couldn't get it working at the time. But I no longer remember exactly what was the issue. I should take another look.

I'll merge in a sec. I think we can get rid of the constructor's completely, instead of `= default`.

jonesmz:
RE: Prettier. Yup, glad to help. I'm learning a lot about how game engines work. Great learning experience.

RE: Caching. Sure, makes sense.

RE: Dropping constructors. Sure, that'd be fine w/ me. I normally specify all of the constructors / assignment operator, even if with = default, but that's because a lot of the guys I work with are still in the C++98 mindset, and they sometime struggle to handle post C++11 code, so I do that to make it clear what's happening for their sake.

Documentation/Manuals/User/serializingObjects.md
```diff
@@ -145,15 +145,13 @@ public:
...
-	MyComponentRTTI()
-		:mInitMembers(this)
-	{ }
+	MyComponentRTTI() = default;
 	
 	// ... 
 };
 ~~~~~~~~~~~~~
 
-Field definition portion of the RTTI type always begins with the @ref bs::BS_BEGIN_RTTI_MEMBERS "BS_BEGIN_RTTI_MEMBERS" macro, and ends with the @ref bs::BS_END_RTTI_MEMBERS "BS_END_RTTI_MEMBERS". If field definition is provided you must also provide a constructor that initializes the *mInitMembers* with the *this* pointer.
```
BearishSun: Nice catch fixing the docs, I would have forgotten that.

jonesmz: I just ran "grep -I -e mInitMembers -r .", so that caught it. I would have forgotten as well.

Though, it wouldn't have caused any harm to initialize in the constructor. Just would have been annoying to users of the API.

Source/Foundation/bsfUtility/Reflection/BsRTTIType.h
```diff
@@ -70,211 +70,211 @@ namespace bs
...
-		META_InitAllMembers(MyType* owner)														\
+		META_InitAllMembers(MyType* owner)																		\
 		{																						\
-			owner->META_InitPrevEntry(META_LastEntry());										\
+			owner->META_InitPrevEntry(META_LastEntry());																\
 		}																						\
 	};																							\
 																								\
-	META_InitAllMembers mInitMembers;
+	META_InitAllMembers mInitMembers{this};
```
BearishSun: I think the only real change in this file is this line, and the comment at the top. The other changes seem to mess up alignment of the backslashes in macros, so I just reverted the old file and applied those changes manually, to avoid fixing all of them individually. Hopefully I didn't miss any other actual changes.

jonesmz: Hrmm, well, actually I made all of those changes specifically because in my editor (I was using the "nano" command line editor), the back slashes were significantly mis-aligned.

I mean, each editor renders tabs differently, but I manually counted the tabs on a couple of lines, and came away thinking they weren't aligned based on the number of tab characters in the file. 

I don't care one way or another though, so whatever you want to do is fine with me.

Though, note that there were numerous lines below the macros that had extraneous spaces at the end of the line.

BearishSun: If you can set up your tabs to be 4 spaces (or ideally, if nano can handle reading .editorconfig file) I think everything should be aligned (at least it is when I'm viewing it in VS).

BearishSun:
Merged.

PR #86, opened 2018-04-22
Use resize instead of reserve on String to avoid index out of range e…

jayrulez:
…xception when terminating string.


BearishSun:
Thanks, this has been merged with b2eddbed901eed7a7f14bb30c7c20b9eec87df10. I've also removed `buf[32] = '\0'` line, as it was writing out of range, and also was redundant.

PR #93, opened 2018-05-05
Remove duplicated mFrustumData clear in `void DrawHelper::clear()`

nemerle:
I've looked `DrawHelper` over, and it does not look like any of the member `Vector` instances was skipped over, so `mFrustumData.clear()` was just redundant.


BearishSun:
Indeed, it was totally redundant. Merged.

PR #94, opened 2018-05-05
Fix an error in Win32GLSupport::selectPixelFormat

nemerle:
Just a missing `{` `}` that was causing spurious `attribList.push_back(GL_TRUE);` when `WGL_EXT_framebuffer_sRGB` was not present.


BearishSun:
Surprised this didn't cause more issues. Merged.

PR #95, opened 2018-05-05
Fix normalsFromSmoothing

nemerle:
The wrong index was used in the second loop of normalsFromSmoothing


BearishSun:
Thanks, merged.

PR #98, opened 2018-05-07
Fix Animation::blend2D

nemerle:
Wrong AnimationClipInfos were referenced.

File was edited in qtcreator, so unused end-of-line spaces were removed, I hope it's not a big problem.


BearishSun:
Looks good, merged. Unused end-of-line spaces are fine to remove. Makes it a bit harder to read the change list but ultimately they are better off gone.

PR #100, opened 2018-05-07
Use `bs_new` in Any::Data::clone

### LINKED ISSUE #99: BsAny new/free issue
The `Any::DataBase::clone()` is allocating memory with `new Data`, but `~Any` is using `bs_delete` to free that memory. Shouldn't the `Any::DataBase::clone()` use `bs_new` same as `Any(const ValueType& value) ` constructor ?

nemerle:
This normalizes the allocator usage.

closes #99


BearishSun:
Ah sorry, just saw this PR as I was pushing a fix.

PR #102, opened 2018-05-07
Various memory fixes

nemerle:


Source/Foundation/bsfCore/Mesh/BsMesh.cpp
```diff
@@ -324,7 +324,7 @@ namespace bs
 
 		mIndexBuffer = IndexBuffer::create(ibDesc, mDeviceMask);
 
-		mVertexData = SPtr<VertexData>(bs_new<VertexData>());
+		mVertexData = bs_shared_ptr<VertexData>(bs_new<VertexData>());
```
jonesmz: Could you elaborate on the benefit here?

Source/Foundation/bsfCore/Mesh/BsMesh.cpp
```diff
@@ -637,7 +637,7 @@ namespace bs
 		desc.subMeshes.push_back(SubMesh(0, initialMeshData->getNumIndices(), drawOp));
 		desc.usage = usage;
 
-		SPtr<Mesh> mesh = 
+		SPtr<Mesh> mesh =
 			bs_shared_ptr<Mesh>(new (bs_alloc<Mesh>()) Mesh(initialMeshData, desc, deviceMask));
```
jonesmz: Shouldn't this be "bs_new" instead of "bs_alloc" ?

Source/Plugins/bsfOpenAudio/BsOAAudio.cpp
```diff
@@ -12,10 +12,10 @@
 namespace bs
 {
 	OAAudio::OAAudio()
-		:mVolume(1.0f), mIsPaused(false)
+		:mVolume(1.0f), mIsPaused(false), mDevice(nullptr)
```
jonesmz: Could you do this as default values in the class declaration? That would allow the constructor to be marked as "= default;"

Source/Foundation/bsfCore/Mesh/BsMesh.cpp
nemerle: The code at this point is simply incorrect:
If at any point in time `GenAlloc` allocator begins to do clever things, the code 'will' break.
When mVertexData is released/freed currently it simply calls delete on the stored pointer, which was allocated by bs_new ( and thus using GenAlloc), and thus needs to be freed using  `bs_delete`

Source/Foundation/bsfCore/Mesh/BsMesh.cpp
nemerle: While I do agree that this could be simplified by using bs_new, I do not feel comfortable enough with the codebase yet, to authoritatively state that it should be done. I think @BearishSun should take a look at this, and tell us why we should be using `placement new` + `bs_alloc` combo at this point :smile:

Source/Plugins/bsfOpenAudio/BsOAAudio.cpp
nemerle: Oh I agree that making those into default values in class declaration could be done. But from what I've seen of the bsf internals, this particular c++11 feature is mostly used for structs but not for classes ? I've tried to maintain consistency with the rest of bsf codebase.

Source/Foundation/bsfCore/Mesh/BsMesh.cpp
nemerle: Well, the whole bsf codebase has over 40 places where `new (bs_alloc<...>)` construct is in use. I can change all of those, but prefer to reduce the amount of work in this PR to what it meant to be, a few memory related fixes :)

jonesmz: Of course, I understand.

Unless @BearishSun has an objection, I don't see why moving those 40+ places to bs_new would be bad.

jonesmz: I mean, in a separate pull request, possibly not even in a single one.

nemerle: Agreed :smile:

BearishSun: Separate PR for this sounds good. The reason I used placement new is that `Mesh` and various other classes with `create()` methods have non-public constructors. This used to mean that `bs_new` couldn't call them, and the code failed compiling on earlier versions of MSVC. But it appears to be working now with MSVC and GCC, so I don't mind changing it (preferably to `bs_shared_ptr_new`). I'm not sure what the standard says about this however, a global method calling a private constructor.

Source/Plugins/bsfOpenAudio/BsOAAudio.cpp
BearishSun: Ultimately I'd like to move all code to use in-class initializers. Constructor initializer lists were the old way of doing things. Note that I would prefer not to use `= default` constructor if the constructor itself contains no code and there are no other constructors. Preferably we just get rid of the constructor altogether.

BearishSun:
Thanks, merged. Made some minor formatting changes and moved OAAudio member initialization to class.

nemerle:
It seems one of the fixes did not make it, the one for linux `Platform::_startUp()`, was there something wrong with it ?

BearishSun:
It was merged (Your link shows it in the commit log).

nemerle:
Right You are, my git got a bit mangled, sorry for the confusion

PR #103, opened 2018-05-08
Make EnumClassHash operator() constexpr

jonesmz:
It's just a static_cast, no reason it shouldn't be constexpr.

By the way, I notice that there's no typedef for std::forward_list. It has slightly better performance than std::list for certain operations (and worse for others, of course).


BearishSun:
Thanks, merged. Feel free to add forward_list if you feel like it. If not I'll keep it in mind for when it seems like it could come in handy. Maybe it would be good to make the default `List` be a forward list, since it feels like it's more commonly required.

PR #105, opened 2018-05-08
Bugfix: improperly locked mutexes.

nemerle:
`Lock(mMutex);` is actually defining a local `Lock` variable named `mMutex` and not locking the mMutex

Since `-Wno-unused-variable` was added to gcc/clang options they were not reporting this.

Thanks go to gcc's 8.1 new `-Wparentheses` warning

### DIFF
--- a/Source/Foundation/bsfCore/Managers/BsGpuProgramManager.cpp
+++ b/Source/Foundation/bsfCore/Managers/BsGpuProgramManager.cpp
@@ -3,7 +3,7 @@
 #include "Managers/BsGpuProgramManager.h"
 #include "RenderAPI/BsRenderAPI.h"
 
-namespace bs 
+namespace bs
 {
 	SPtr<GpuProgram> GpuProgramManager::create(const GPU_PROGRAM_DESC& desc)
 	{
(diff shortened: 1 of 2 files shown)


Source/Foundation/bsfCore/Managers/BsGpuProgramManager.cpp
```diff
@@ -96,14 +96,14 @@ namespace bs
 
 	void GpuProgramManager::addFactory(const String& language, GpuProgramFactory* factory)
 	{
-		Lock(mMutex);
```
jonesmz: I didn't realize that that syntax would set a new variable. How archaic!

I had assumed that was a macro, for some reason.

jonesmz:
Maybe we should remove no-unused-variable from the list of warning flags?

Source/Foundation/bsfCore/Managers/BsGpuProgramManager.cpp
nemerle: at the first glance it looks like a `Lock` function call, doesn't it ? :smile:

BearishSun:
Nice catch. This isn't the first time I made this mistake so enabling no-unused-variable is probably a good idea. As long as we get rid of all the other code that trigger it (IIRC it was disabled because it triggered at some points were unused variables were due to preprocessor conditionals and such).

BearishSun:
Merged. Note that the convention for local variables is to use camelCase. In particular for locks I tend to just call them `lock`.

PR #108, opened 2018-05-08
Don't make CMAKE_BUILD_TYPE invisible in cmake-gui

### LINKED ISSUE #107: c++17 usage in the codebase.
It seems bsf is using a few c++17 features - nested namespaces for example.
Should those things be removed, or should we enable c++1z flag, and remove `-Wno-c++1z-extensions` from Apple xcode9 build ?

nemerle:
Enable c++1z on Apple.
Set CMAKE_CXX_STANDARD_REQUIRED flag to true, to error-out on non c++14 compilers.

closes #107

### DIFF
--- a/Source/CMake/Properties.cmake
+++ b/Source/CMake/Properties.cmake
@@ -4,7 +4,7 @@ if(CMAKE_CONFIGURATION_TYPES) # Multiconfig generator?
 else()
 	if(NOT CMAKE_BUILD_TYPE)
 		message("Defaulting to release build.")
-		set(CMAKE_BUILD_TYPE Release CACHE INTERNAL "")
+		set(CMAKE_BUILD_TYPE Release CACHE STRING "")
 	endif()
 	set_property(CACHE CMAKE_BUILD_TYPE PROPERTY HELPSTRING "Choose the type of build")
 	# Set the valid options for cmake-gui drop-down list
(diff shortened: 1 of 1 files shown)


nemerle:
@BearishSun the code base is using `OptimizedDebug`, which seems to be closely related to cmake's standard `RelWithDebInfo`, out of curiosity what is the difference ?

BearishSun:
Re `OptimizedDebug`: I believe the intent was to use this configuration primarily for debug purposes, so the name suited better (various debug checks remain enabled, even though its an optimized build). I'm currently rethinking the configurations with the release near, as I want symbols for debugging, yet a build with no debugging checks for the user itself. Might either create a new config, modify OptimizedDebug, or just add debug symbols to Release (the last option seeming like the most likely).

nemerle:
Apparently that's what `RelWithDebInfo` is for :)

Source/CMake/Properties.cmake
```diff
@@ -54,7 +55,7 @@ if(MSVC)
 
 	# Compiler
 	set(BS_COMPILER_FLAGS_COMMON "/GS- /W3 /GR- /WX- /MP /nologo /bigobj /wd\"4577\" /wd\"4530\"")
-	set(CMAKE_CXX_FLAGS "/DWIN32 /D_WINDOWS")
+	add_definitions(-DWIN32 -D_WINDOWS)
```
BearishSun: This line appears to be causing an internal compiler error MSVC, along with other weirdness. Not sure why. I'll switch it back and merge without this change.

nemerle: That's amazing in a way, that different way of adding defines crashes msvc :D

BearishSun: My thoughts exactly :/

BearishSun:
Merged.

PR #109, opened 2018-05-08
Add all of the comparison operators to NativePtr so it can be used in…

jonesmz:
… std::containers more easily


BearishSun:
Thanks, merged.

PR #112, opened 2018-05-09
Fix: BsDynLibManager "unload"

### LINKED ISSUE #111: the function “unload” in BsDynLibManager.cpp doesn't work properly
const auto& iterFind = mLoadedLibraries.find(lib->getName());
	if (iterFind != mLoadedLibraries.end())
	{
		mLoadedLibraries.erase(iterFind);
	}

the 'iterFind' always point to the end,so mLoadedLibraries.erase(iterFind) has no chance to execute！

nemerle:
Since default comparator compares pointers and the `mLoadedLibraries`
needs to compare by `getName()` strings, using template specialization
should solve the problem.
In my limited testing this seems to work correctly, but @XMatrixXiang please check if this solves the issue you were having.
Fixes #111

### DIFF
--- a/Source/Foundation/bsfUtility/Prerequisites/BsStdHeaders.h
+++ b/Source/Foundation/bsfUtility/Prerequisites/BsStdHeaders.h
@@ -223,12 +223,12 @@ namespace bs
 		constexpr T& operator*() const { return *mPtr; }
 		constexpr T* operator->() const { return mPtr; }
 		constexpr T* get() const { return mPtr; }
-		constexpr bool operator< (const NativePtr& rhs) const { return mPtr <  rhs.mPtr; }
-		constexpr bool operator> (const NativePtr& rhs) const { return mPtr >  rhs.mPtr; }
-		constexpr bool operator<=(const NativePtr& rhs) const { return mPtr <= rhs.mPtr; }
-		constexpr bool operator>=(const NativePtr& rhs) const { return mPtr >= rhs.mPtr; }
-		constexpr bool operator==(const NativePtr& rhs) const { return mPtr == rhs.mPtr; }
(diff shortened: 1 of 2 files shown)


XMatrixXiang:
@nemerle Thanks, I'll take a look as soon as I get a chance.

BearishSun:
Thanks for the fix, the issue appears to be resolved. Merged.

Source/Foundation/bsfUtility/Prerequisites/BsStdHeaders.h
```diff
@@ -223,12 +223,12 @@ namespace bs
 		constexpr T& operator*() const { return *mPtr; }
 		constexpr T* operator->() const { return mPtr; }
 		constexpr T* get() const { return mPtr; }
-		constexpr bool operator< (const NativePtr& rhs) const { return mPtr <  rhs.mPtr; }
-		constexpr bool operator> (const NativePtr& rhs) const { return mPtr >  rhs.mPtr; }
-		constexpr bool operator<=(const NativePtr& rhs) const { return mPtr <= rhs.mPtr; }
-		constexpr bool operator>=(const NativePtr& rhs) const { return mPtr >= rhs.mPtr; }
-		constexpr bool operator==(const NativePtr& rhs) const { return mPtr == rhs.mPtr; }
-		constexpr bool operator!=(const NativePtr& rhs) const { return mPtr != rhs.mPtr; }
+		bool operator< (const NativePtr& rhs) const { return mPtr <  rhs.mPtr; }
```
jonesmz: Hold on. Why did you need to remove the constexpr?

nemerle: That's the unfortunate consequence of having to specialize the template, since specialization had to be calling non-constexpr functions, it couldn't be constexpr itself :disappointed:

Issue #115, opened 2018-05-10
GUIInputBox character count is never updated outside of setText method


jayrulez:
This causes various issues using GUIInputBox.

E.g: If you set text to "Hello" using the setText method then you cannot delete (or backspace) and character at index outside the range of the original length.

jayrulez:
While this fixes the issue, I would prefer a more elegant method where setText is always called to update the text so we do not have to worry about missing updating the character count in other places.

@BearishSun let me know what you think.

BearishSun:
I wouldn't mind a more elegant solution if you feel like it (perhaps a simpler, internal version of `setText`).

PR #116, opened 2018-05-10
Fixes GUIInputBox issues

jayrulez:
Fixes issue #115


jonesmz:
Could you split this PR out into a different branch that has fewer Intermediate commits? Merging this would add a lot of stuff to the git log.

PR #118, opened 2018-05-11
GUIInputBox: Ensure mNumChars is always updated when text is modified

jayrulez:
Fixes issue #115


jayrulez:
@jonesmz

BearishSun:
Thanks, merged.

PR #119, opened 2018-05-11
Remove many unused assignments + memset fix

nemerle:
Clang static analyzer was complaining about those:
* assigning values to `memory = ` and not using them at all
* possible memset with nullptr, and size 0, just to be on the safe side
moved them inside the corresponding `if`


BearishSun:
Looks good, merged.

PR #120, opened 2018-05-11
Add noreturn attribute to PlatformUtility::terminate

nemerle:
Should help out with optimization when compiling without exceptions and
BS_EXCEPT is called.

Also helps with static code analyzers.


BearishSun:
Looks good, merged.

PR #121, opened 2018-05-11
Mem fixes

nemerle:
* 2 memory leaks, 1 is pretty rare, the other would happen on every call of getTempDirectoryPath
* fix invalid component access in flipComponentOrder

### DIFF
--- a/Source/Foundation/bsfCore/Image/BsPixelUtil.cpp
+++ b/Source/Foundation/bsfCore/Image/BsPixelUtil.cpp
@@ -8,16 +8,16 @@
 #include "Image/BsTexture.h"
 #include <nvtt.h>
 
-namespace bs 
+namespace bs
 {
 	/**
 	 * Performs pixel data resampling using the point filter (nearest neighbor). Does not perform format conversions.
 	 *
 	 * @tparam elementSize	Size of a single pixel in bytes.
 	 */
-	template<UINT32 elementSize> struct NearestResampler 
+	template<UINT32 elementSize> struct NearestResampler
 	{
-		static void scale(const PixelData& source, const PixelData& dest) 
+		static void scale(const PixelData& source, const PixelData& dest)
 		{
 			UINT8* sourceData = source.getData();
 			UINT8* destPtr = dest.getData();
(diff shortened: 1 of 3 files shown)


Source/Foundation/bsfUtility/FileSystem/BsDataStream.cpp
```diff
@@ -94,7 +94,7 @@ namespace bs
 		// Read the entire buffer - ideally in one read, but if the size of
 		// the buffer is unknown, do multiple fixed size reads.
 		size_t bufSize = (mSize > 0 ? mSize : 4096);
-		std::stringstream::char_type* tempBuffer = (std::stringstream::char_type*)bs_alloc((UINT32)bufSize);
+		std::array<std::stringstream::char_type,4096> tempBuffer;
```
jonesmz: It might be *slightly* gentler on stack usage to instead use alloca(), but there's probably no major benefit to doing so.

Good find :-)

Source/Foundation/bsfUtility/Private/Unix/BsUnixFileSystem.cpp
```diff
@@ -336,12 +336,12 @@ namespace bs
 		}
 
 		tmpdir.append("/banshee-XXXXXX");
-		size_t len = tmpdir.size()+1;
-		char *nameTemplate = bs_newN<char>(len);
-		snprintf(nameTemplate, len, "%s", tmpdir.c_str());
+		Vector<char> nameTemplate(tmpdir.data(),tmpdir.data()+tmpdir.size());
+		nameTemplate.push_back(0); // null terminate the name template.
```
jonesmz: You're doing two allocations here.

One for the vector itself, and a second when you call push_back().

Instead, create an empty vector (or string, since it's a vector of char, std::string would be more directly analogous), and reserve the total size you want to use. Then push_back().

Source/Foundation/bsfCore/Image/BsPixelUtil.cpp
```diff
@@ -2034,12 +2034,12 @@ namespace bs
 
 		// Ensure unused components are at the end, after sort
 		if (pfd.componentCount < 4)
-			compData[4].shift = 0xFF;
+			compData[3].shift = 0xFF;
```
jonesmz: Write paste end error?

nemerle: well, both of those looked like off-by-one errors to me :)

Source/Foundation/bsfUtility/FileSystem/BsDataStream.cpp
nemerle: Yeah, 4kb for a function that is not called recursively should not blow up the stack 😄

Source/Foundation/bsfUtility/Private/Unix/BsUnixFileSystem.cpp
nemerle: You're right, the size of nameTemplate will be exactly equal to tmpdir.size(), I'll replace it with an one-liner.

jonesmz:
I meant "was this a correction of an off by one error". Sounds like yes.

BearishSun:
Looking good, thanks. Merged.

PR #123, opened 2018-05-13
Make NativePtrs comparison operators live outside the class, so that they can be constexpr.

jonesmz:
Lets try this rodeo one more time :-)

Now NativePtr's comparison operators aren't member functions, freeing up template specializations from forcing the default implementation to be non-constexpr.

Until Visual Studio's implementation of std::set supports UniquePtr (which, btw, the most recent update may have fixed that, and now that VS2015 isn't supported anymore, I'll investigate that again), this should be the final word on the whole dynlib situation.


BearishSun:
Looks like the best option, thanks. Merged.

PR #125, opened 2018-05-15
Prevent copying of BsModule by default

jonesmz:
It doesn't make sense to allow for Module to be copied, or moved, really, as it represents a singleton instance. Copying or moving such an object would ultimately end up pointing at the same instance.


BearishSun:
Makes sense, thanks. Merged.

PR #126, opened 2018-05-16
Use UPtr to manage DynLib lifetimes inside of BsDynLibManager.

jonesmz:
The problem that was causing this to work in Linux but not VS2017 is two fold.

1) Gcc defines the copy constructor for std::set as "default", which allows for them to be disabled in situations where the items inside the set aren't copyable, but VS2017 doesn't. They explicitly define them.

2) BsModules was set to have a default copy constructor, which meant that std::set had it's copy-constructor explicitly used in that default copy-constructor.

By disabling the copy and move constructors for BsModules, we eliminate half of the one-two-punch.

So now we can use UniquePtr to manage the lifetime of loaded DynLib objects.

### DIFF
--- a/Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
+++ b/Source/Foundation/bsfUtility/Utility/BsDynLibManager.cpp
@@ -5,16 +5,27 @@
 
 namespace bs
 {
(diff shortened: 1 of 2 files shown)


BearishSun:
Nice working figuring it out :) Merged

PR #142, opened 2018-06-04
Full Cotire Support

### LINKED ISSUE #139: Cotire and Precompiled Headers
As I'm starting to work to integrate with bsf, I think having precompiled headers will help with compile times significantly. The idiomatic way to do this with CMake is to use cotire. I've been doing some poking around and it seems perfectly possible to do. If I were to make those changes in a more proper way would you be interested in having them? I'm thinking a setup where if it detects cotire it uses it, but if it doesn't, it just pretends like nothing happened.

cwfitzgerald:
This set of patches enables full cotire support. There are sill problems with building with cotire on mac, which are still to be fixed, but I am unable to fix them due to me not having a mac at all.

This also enabled `-fdiagnotisics-color=always` on ninja builds. 

This resolves #139.


Source/CMakeLists.txt
```diff
@@ -82,6 +82,10 @@ set(PHYSICS_MODULE_LIB bsfPhysX)
 configure_file("${BSF_SOURCE_DIR}/CMake/BsEngineConfig.h.in" "${BSF_SOURCE_DIR}/Foundation/bsfEngine/BsEngineConfig.h")
 configure_file("${BSF_SOURCE_DIR}/CMake/BsFrameworkConfig.h.in" "${BSF_SOURCE_DIR}/Foundation/bsfUtility/BsFrameworkConfig.h")
 
+## Set files disallowed in precompiled headers
+set_directory_properties(PROPERTIES
+    COTIRE_PREFIX_HEADER_IGNORE_PATH "${CMAKE_CURRENT_SOURCE_DIR};/usr/include/clang")
```
BearishSun: Is this still required since we have more robust search for `inttypes.h`?

cwfitzgerald:
Nope, just forgot to remove it. I'll remove it when I get to work. 
---
Connor Fitzgerald

Source/Foundation/bsfCore/Renderer/BsCamera.cpp
```diff
@@ -14,6 +14,9 @@
 #include "Renderer/BsRenderer.h"
 #include "Scene/BsSceneManager.h"
 
+// Do not delete this, it prevents massive problems with windows headers across the codebase
+#include "Private/Win32/BsWin32Undefs.h"
```
BearishSun: Having a private header in a public header will not work since private headers don't get installed, so it will break in binary distributions. Also, I'd prefer to move this to a more generic location - perhaps even replacing all usages of "windows.h" with a custom header that automatically undefines the problematic macros?

Source/Foundation/bsfCore/CMakeSources.cmake
```diff
@@ -621,16 +621,31 @@ set(BS_CORE_INC_PLATFORM_LINUX
...
 	"bsfCore/Private/Linux/BsLinuxGamepad.cpp"
-	"bsfCore/Private/Linux/BsLinuxMouse.cpp"
-	"bsfCore/Private/Linux/BsLinuxKeyboard.cpp"
+	"bsfCore/Private/Linux/BsLinuxFolderMonitor.cpp"
 )
 
+# Force unity builds to isolate these files into their own section. 
+# COTIRE_START_NEW_UNITY_SOURCE causes the source file to be the first one included in a new file. 
+# This isolates the X11 headers which define inane things.
+set_source_files_properties ("bsfCore/Private/Linux/BsLinuxPlatform.cpp" PROPERTIES
```
BearishSun: Could this be something we do for Windows as well, to avoid the issues with "windows.h"? If so could we just automate the process and have cmake iterate over `BS_CORE_SRC_PLATFORM_LINUX` (and other platform-specific) files and ignore them all? I don't mind those files being excluded from the unity builds (and I'm assuming their headers are also excluded from PCH), if it means for a cleaner code without duplication.

BearishSun:
Thanks for the PR, look good aside from the stuff I commented on. I'll take a look at what's wrong with macOS as soon as I get a chance.

Btw, is there a way to disable cotire messages in CMake log (e.g. cotire was enabled for target ...)?

Also, I don't see `cotire.cmake` included in the PR, but it is present in the examples PR. Was it disabled for the plain library build for a reason or?

Source/Foundation/bsfCore/Renderer/BsCamera.cpp
cwfitzgerald: Isn't this a public source file anyway?

Either way, I agree that replacing windows.h with a proxy header would be better. I'll get on that.

BearishSun: Ah sorry, I thought this was `BsCamera.h`, my bad.

Source/Foundation/bsfCore/CMakeSources.cmake
cwfitzgerald: Yes, I haven't tested unity builds on windows yet as I didn't get as much time on my desktop as I had hoped. 

The reason I am hesitant to just mark this so cotire ignores it is that it would illuminate some of the purpose of the unity builds. By keeping all the X11 comparable headers in one place they get a single unity build together causing at least some level of unity-ness.

Source/Foundation/bsfCore/Renderer/BsCamera.cpp
cwfitzgerald: Cotire only walks source files to build the PCH list, (though it gets transitively included ones) so the best way (besides doing what you suggested) is to have it included directly in a source file.

Source/Foundation/bsfCore/CMakeSources.cmake
BearishSun: I'm worried about things breaking in weird ways for someone unfamiliar with the codebase. 

i.e. just including a X11 header in some other Linux specific file will break the build (I'm assuming) - and likely in not a very obvious way. I don't like having such hidden requirements in the codebase, if there's a way to avoid them.

cwfitzgerald:
There doesn't seem to be a way to disable cotire messages, but if we're going to host cotire ourselves, we can just remove the message.

The reason I didn't include cotire in the library is that I thought you would want to have the default build not change, but have the option open if you could (it's less aggressive for me to say "hey you need to use this thing in your build" vs "hey would it be possible for me to put a thing to _optionally_ use this thing in your build"). But if it is okay with you, I will go ahead and add cotire to this as well.

Source/Foundation/bsfCore/CMakeSources.cmake
cwfitzgerald: I'm torn, as the X11 header is possibly the worst header I've ever seen besides <windows.h> so I want to shun it's usage as much as possible, but I agree that we shouldn't leave the build breaking randomly. 

I think I will let these ones out of the unity build.

BearishSun:
Having the ability do disable it would be nice (at least until we guarantee it works perfectly for everyone - this way I could even merge the PR earlier). So we include cotire, but add a `ENABLE_COTIRE` CMake option, disabled by default (for now)?

cwfitzgerald:
Sounds good to me.

cwfitzgerald:
Should be good now minus issues with windows unity builds and mac everything.

cwfitzgerald:
There's issues with the windows.h solution....

BearishSun:
I've taken a quick look. One of the issues is simple, `BsWindows.h` includes itself instead of `<Windows.h>`.

But another issue is that undefining `near` and `far` seems to screw up some other Windows headers. I'd be up for completely excluding `Windows.h` from PCH, but I haven't yet found a good way of locating it.

cwfitzgerald:
There's already a function that locates windows.h, you can just ask cmake to find it. I would like to keep Windows.h (and it's giant size) in the PCH if possible. I'm going to be home and can actually debug it on windows in a bit.

BearishSun:
That function doesn't find the header, at least not on my end.

I do wonder if `Windows.h` maybe even slows down the build ultimately, because of its giant size? Since we are including it in just a few files without PCH, but with PCH we are including it in every single file. Even in its precompiled format, it might be ultimately slowing things down. But I haven't profiled it.

BearishSun:
While we're at it, some other Windows headers are only included in one single file (or perhaps a couple). Not sure if cotire detects that and can exclude it from PCH in that case? Would be nice to disable those as well and test build times.

cwfitzgerald:
I'll see what I can do about it, it will probably require another workaround like we did for clang. As for the other headers I can look into the common ones which are isolated in a couple files and detect them too. 
---
Connor Fitzgerald

cwfitzgerald:
The problem is that right now I don't have a reliable way of detecting where windows.h is, as it works on my windows 10 machine. I may be able to try something with /showIncludes and parsing the output of that using good old regexes. I'm not sure exactly which ones include `windows.h` by transitive dependencies.

cwfitzgerald:
Alright, I think I have everything down... except getting cotire to actually leave the header out. I am still fighting with it to see what the correct combination of slashes or whatever to get cotire to realize that we're talking about the same header.

BearishSun:
Such a huge hassle for a single header :(

I did notice that cotire added a full path to `windows.h` in the precompiled header file, so I'm guessing it has some internal mechanism for locating it? Might be worth trying to find it. Although you seem to be nearly there anyway.

Another idea I had was to modify cotire so we can filter includes based on their name, instead of the full path.

cwfitzgerald:
FOUND THE PROBLEM! Something with the whole X11 bit being on windows too was causing it to fuck up. I'm going to see what the final resolution of all this will be.

cwfitzgerald:
I was wrong about the first solution, turns out I was searching for `windows.h`, finding it, then trying to get it to block the inclusion of `Windows.h`. CMake is case sensitive.... currently works, but the compiler can't find the .dll's it needs.

cwfitzgerald:
Alright, that last push did it. I figured out all the problems (I hope). As of right now, the working/non working feature list is as follows

Working:
- Windows PCH
- Linux PCH
- Linux Unity

Not Working
- Windows Unity
- Mac PCH

Untested
- Mac Unity

I'm completely burnt out of staring at cmake problems and fighting with visual studio. It's good enough for my usecase right now, and I'll probably come back to this in a different PR to make cotire a more robust feature. I definitely don't think it should be advertised in docs until we know it's working right.

BearishSun:
Sure, thanks for all the work. I'll merge when I get a moment to properly review everything so I don't rush it, sometimes this week. Then take a look at the mac stuff when I get more time.

BearishSun:
Merged, f80c1c53ac0b4fb82b0670b47857c6e0cc5f1160. Thanks again.

PR #145, opened 2018-06-09
Make binary dependency website able to be changed

cwfitzgerald:
As I'm integrating bsf into my project, I realize that it would be kind of unfair to expect your server to deal with each of my many builds downloading all the BSF files (and gcc and binutils) every single time (8 builds, possibly 10 ci instances/day, 500Mb each). As a solution, I set up my server as a caching reverse proxy to yours, so I can point stock bsf to my server and it will cache the downloads so it doesn't constantly be hit by building. This patch allows you to change the url it searches for files in.


BearishSun:
Thanks for thinking of the bandwidth! Looks good, merged.

Issue #149, opened 2018-06-12
Isolate CMake Flags for Parent Projects


cwfitzgerald:
When I add bsf in my project, it overwrites my flags and cause issues (I use exceptions and RTTI) as well as prevent me from getting some warnings. The flags need to be managed in a better way so flags are carried as needed (internal flags kept internal and external flags passed along). Changing to "linking" against targets that have the proper flags are the best and most modern solution to this.

I'll submit a PR once I've got a minute.

BearishSun:
Yeah that's a good point, I haven't considered it until now. Adding the flags to targets is definitely the way to go.

cwfitzgerald:
The Debug/OptimizedDebug/Release pattern is also causing problems as almost no one will expect you to change the defaults, and will cause a bunch of funky things to happen. The normal pair is `Debug/RelWithDebInfo/MinSizeRel/Release` which works fine with other libraries. I will probably split things up.

cwfitzgerald:
As soon as I type this I realize why it's causing problems. Because I am including properties, it is changing CMAKE_CXX_FLAGS etc in my scope not yours (which is created with CMAKE_CXX_FLAGS). Investigating...

BearishSun:
Closing this manually since #150 should have closed it automatically.

PR #150, opened 2018-06-13
Fix Build Issues with Library as Submodule

cwfitzgerald:
"Fixes" #149, and by fixes I mean does just enough to work well.


BearishSun:
Oh I just really hate that original CMake config names. Why have both RelWithDebInfo and Release? It's not like anyone sane should build release software without at least keeping debug symbols in-house...

But oh well, if it causes compatibility problems with external libraries I guess that's a small price to pay.

Anyway, thanks, merged :)

cwfitzgerald:
I know that until I started to think about actual distribution (running the code on another machine) the idea of debug symbols for release mode seemed silly (why would you need those if you only debug in debug mode :P) and traditionally, most linux based software comes without debug symbols, with the assumption that if something went wrong you'd just compile it yourself in debug mode. As cmake is essentially based around the linux compilation model, it makes sense why they chose the defaults as they did, even though for production end-user software it doesn't make any sense.

I'm going to submit another PR, as bsf still clobbers flags that I need to pass to clang to make the binutils work. Waiting on travis-ci to get back to me to see if I broke anything too badly.

PR #151, opened 2018-06-14
Remove redundant check in PrefabDiff::generateDiff

nemerle:
Since output is uninitialized, the comparison `if (output == nullptr)` will always be true

### DIFF
--- a/Source/Foundation/bsfCore/Scene/BsPrefabDiff.cpp
+++ b/Source/Foundation/bsfCore/Scene/BsPrefabDiff.cpp
@@ -39,7 +39,7 @@ namespace bs
 		// would be more efficient to do it once outside of this method. I'm keeping it this way for simplicity for now.
 
 		// Rename instance objects so they share the same IDs as the prefab objects (if they link IDs match). This allows
-		// game object handle diff to work properly, because otherwise handles that point to same objects would be 
+		// game object handle diff to work properly, because otherwise handles that point to same objects would be
 		// marked as different because the instance IDs of the two objects don't match (since one is in prefab and one
 		// in instance).
 		Vector<RenamedGameObject> renamedObjects;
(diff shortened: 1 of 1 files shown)


BearishSun:
I believe I kept it on purpose. If someone modifies that code by adding a new field to compare, he also needs to remember to re-add that check (if his code comes earlier), which I think is an easy thing to forget.

nemerle:
That's a fair point. So how about this approach ?
We create `PrefabObjectDiff output` on stack, and mark it as not valid.
Every check afterwards accesses already valid stack allocated diff, and marks the diff as valid
At the end of that function we `move` the stack allocated object into an heap allocated instance, thus hopefully re-using already allocated `Vector` memory ?

BearishSun:
Hmm. We are now creating a `PrefabObjectDiff` every time that method is called, whether it's needed or not. Sure it's on the stack, which saves a dynamic memory allocation, but all the constructors for all the types it contains still need to be called, and their data initialized. And stuff like `std::vector` or `std::string` don't guarantee it won't dynamically allocate something in its constructor either (although most implementations wont).

And then at the end we also need to pay the cost of copying everything back. Again we save on copying dynamic memory since we're using std::move, but there's a lot of non-movable contents that still need to be copied.

Branches should be considerably cheaper than copying any kind of memory around, so I feel it's a hefty cost to pay to eliminate a few branches.

nemerle:
I'll try to verify this on godbolt, see overhead this vs multiple sharedptr constructor calls .

nemerle:
https://godbolt.org/g/XwSzCN
Instruction counts  with `-O3 -fno-rtti  -fno-exceptions`
```
       'original' -   'new'
CLANG 6    574    -    437
GCC 8.1    623    -    778
VS17      2543    -   2900
```
You're right, the original code is also faster according to http://quick-bench.com/sr4MqfrAnyQLRh85zk0g037WZag

BearishSun:
Thanks for the detailed analysis. Interesting result with Clang - I'm guessing it sees more optimization opportunities with the stack allocated version?

nemerle:
Yup, the ast is a bit simpler with stack allocated var.

It would be interesting to see how the assembly looks like with EASTL, since their containers are guaranteed not to allocate heap until first push/insert/emplace is done.

PR #153, opened 2018-06-14
Thread pool cleanup

nemerle:
Not sure if `static constexpr int UNUSED_CHECK_PERIOD = 32;` should be public, probably not :)


BearishSun:
Looks good, thanks, merged. Yep I made the constant local to the implementation file.

PR #154, opened 2018-06-14
Prevent BSF from Eating Parent Project Flags (extension of #150)

cwfitzgerald:
This is a small change to prevent the flags from getting overwritten. It is generally considered the parent project's responsibility to make sure sub directory inclusion happens in an order that makes sense and doesn't cause over pollution of flags (bsf shouldn't get my `-ffast-math` for example). However, there are flags that are set on the command line that need to be honored. This solves this problem. 

This extends #150.


BearishSun:
Yep that makes sense, thanks, merged.

PR #155, opened 2018-06-14
Bs slfx compiler fix

nemerle:
One typo fixed + simplify parseCodeBlock a little bit.


BearishSun:
Good catch with the sampler state typo, thanks! Merged.

PR #156, opened 2018-06-15
Isolate All Build Artifacts into Build Directory

cwfitzgerald:
## Preface

Note this PR isn't currently intended as a complete implementation ready for merge, but as a way to start discussion about the issues it's trying to solve as well as the code itself. This hasn't been tested nearly enough on any platform, or in any kind of other scenario.

## What

Pulling all the downloading and configured dependencies into the build folder as opposed to the source folder.

## Rationale

1. I have encountered scenarios where there are multiple builds for multiple platforms pulling form the same source directory (shared dir between a windows host and linux vm) causing conflicts in the dependency resolution.
2. CLion initializes all builds at the same time by calling CMake concurrently. Because of the inherent race condition with downloading and extracting the dependency and asset files, the first couple times I try to reload the project it fails, until something finally wins the fight and things can continue on as normal.
3. To me, it seems like good policy that nothing should change in the source tree as a result of a build happening. This allows many different builds to happen simultaneously with the advantage that if a build directory gets deleted, **everything** gets deleted, and it's as if you just cloned the repository

## Comments

I know for sure I am missing parts, and I haven't done full testing yet in a variety of scenarios, but can you point me in the direction of anything else I may be missing. Breaking the fixed location paradigm within the cmake code is hard as it's hard to tell what everything touches.

### DIFF
--- a/Source/CMake/HelperMethods.cmake
+++ b/Source/CMake/HelperMethods.cmake
@@ -222,8 +222,10 @@ mark_as_advanced(BS_BINARY_DEP_WEBSITE)
 
 function(update_binary_deps DEP_PREFIX DEP_FOLDER DEP_VERSION)
 	# Clean and create a temporary folder
-	execute_process(COMMAND ${CMAKE_COMMAND} -E remove_directory ${PROJECT_SOURCE_DIR}/Temp)	
-	execute_process(COMMAND ${CMAKE_COMMAND} -E make_directory ${PROJECT_SOURCE_DIR}/Temp)	
+	set(DOWNLOAD_TEMP_FOLDER ${PROJECT_BINARY_DIR}/TempDeps)
+
+	file(REMOVE_RECURSE ${DOWNLOAD_TEMP_FOLDER})
+	file(MAKE_DIRECTORY ${DOWNLOAD_TEMP_FOLDER})
 
 	if(WIN32)
 		set(DEP_TYPE VS2015)
@@ -234,7 +236,7 @@ function(update_binary_deps DEP_PREFIX DEP_FOLDER DEP_VERSION)
 	endif()
 
 	set(BINARY_DEPENDENCIES_URL ${BS_BINARY_DEP_WEBSITE}/${DEP_PREFIX}Dependencies_${DEP_TYPE}_Master_${DEP_VERSION}.zip)
-	file(DOWNLOAD ${BINARY_DEPENDENCIES_URL} ${PROJECT_SOURCE_DIR}/Temp/Dependencies.zip 
+	file(DOWNLOAD ${BINARY_DEPENDENCIES_URL} ${DOWNLOAD_TEMP_FOLDER}/Dependencies.zip
 		SHOW_PROGRESS
 		STATUS DOWNLOAD_STATUS)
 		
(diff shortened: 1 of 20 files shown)


BearishSun:
Sounds good to me, ideally we also allow the dependencies to be installed as it is currently since I feel like I sometimes clear the build folder often and would rather not re-download things every time (remains to be seen how much on an issue this really is). But looking at the code, you seem to have covered that.

You'll likely need to add an install step for `BsEngineConfig.h` and `BsFrameworkConfig.h`, since all headers are currently installed by globbing through the `Source` directory, which will miss them now.

Speaking of `BsEngineConfig.h`, it hardcodes the path to the source directory in `RAW_APP_ROOT`. The most important use of that variable is to be able to find the `Data` folder and related assets (in `BsPaths.cpp`), which are expected to be relative to that folder. Since that folder no longer needs to be relative to the source directory that logic will need to change (possibly just by passing a direct path to the Data folder using a different config variable).

Aside from that I can't think of anything, might be some other issues when it comes it including `bsf` in Banshee 3D as there's some special handling for its paths, but I can figure that out when the time comes.

BearishSun:
And perhaps a bigger issue, `Data` folder is partially part of the git repository itself (shaders in particular, since they change often), and I'm not sure I like the idea of having it in two locations.

It might be best to keep the data folder as-is. It's identical for all platforms and would be part of the repository itself if it wasn't for worries about bloating repository size due to binary file versioning.

That would also solve the issue I mentioned above.

I'm not sure how that plays with the CLion issue, but perhaps that can be solved via some kind of a lock file?

cwfitzgerald:
Good points, I agree on all sides. 

If we keep data the same, CMake does allow to put a lock on a folder/file to prevent modification. The docs warn that this may not be honored by other processes, but if it is locked by all functions that mess with the data folder, it should be a non-issue. The second process in on the lock will just spin until data is downloaded, and once it is, it will check and find it is already there. 

The only problem I have faced so far is that there is no real way to tell where the bin dir is to copy binary deps to. I think the solution to this to make a function that can be called by the parent to say "here is my bin folder for X config" and it will copy dependencies as needed. It can be non-re-entrant so the first person to call it gets the bin dir. 

I also need to find out what is moving the location of object files, as once I added bsf, my coverage broke....

w-flo:
The one issue I have with the Data dir is that some generated files come out differently in my environment compared to the environment you (@BearishSun) use when committing those files. So git always tells me I have local changes that I need to commit / stash / drop. It's not a huge inconvenience of course to drop those changes, but they will always come back.

Looking at the diff, one simple step to improve the situation a bit might be to change ```UnorderedSet<String>& includes``` in BsSLFXCompiler.cpp to an ordered set to keep the ordering in ShaderDependencies.json consistent. The differences in .bsl.asset files seem to be a little more complicated. I had to recompile XShaderCompiler because of C++ ABI issues, maybe I used a wrong commit. I can investigate this if there is a goal to have generated files in Data/ be reproducibly byte-identical on all platforms so they can be kept in git.

Also, my flex/bison version is not the same as yours, so I always see a git diff for Source/Plugins/bsfSL/BsLexerFX.{c,h} and Source/Plugins/bsfSL/BsParserFX.{c,h}. I believe you need bison&flex installed on the build system in any case, so maybe those files can be dropped from git and added to gitignore? Or maybe there's a way to create them in the Build directory.

BearishSun:
I do have this issue myself. It seems to be exclusively the result of `BuiltinResources::preprocess` call, which is intended to automatically process changes on built-in assets. 

It might be best to disable this by default, and let users enable it for those that are making changes to built-in resources. Even better it might be best to have this be done by a separate executable which can be ran on demand. This way you won't have framework change resources when you're just trying to run your project.

You can disable lexer/parser generation by disabling `BUILD_BSL` option in CMake. bison/flex is not required by default because it's problematic to install `bison` on mac since it ships with an old version by default. So I decided to provide the files directly instead of generating them locally, and drop the dependencies.

w-flo:
Ah, sounds good (both possible solutions for the assets generation).

I misunderstood the BUILD_BSL flag, and thought it's needed to use the banshee shading language. So I simply disabled it and now it works as expected. :-)

cwfitzgerald:
Sorry this has laid dormant for so long, I've had a lot on my plate. AFAIK all this is waiting for is for me to implement separation Data folder and to get object files in the right place.

BearishSun:
No problem, whenever you get a chance. It will be nice to have these things cleaned up.

cwfitzgerald:
I am going to close this to prevent it from cluttering things up because I no longer have the time to work on it, and I have basically abandoned c++ for my projects. Keep up the good work y'all!

PR #157, opened 2018-06-15
Feature: Cylinder primitive.

ivalduan:
Thanks for this awesome engine. I needed cylinders, so...


BearishSun:
Nice, thanks!

Ideally later on we can merge the cylinder and cone code in `ShapeMeshes3D`, as I feel they should share a lot of common functionality. But it's not important for now.

Merged.

PR #162, opened 2018-06-27
CoreApplication getters for time variables

ivalduan:
I needed public access to some of the variables inside CoreApplication.

Unity scripts have similar functionality public, and I needed them for a custom Leap Motion SDK integration available here (still WIP):
https://github.com/ivalduan/bsfLeap


BearishSun:
Looks like an interesting project!

Before I merge I'm wondering does this perhaps better belong to the `Time` class? Also I wouldn't mind knowing more about exactly in what context is it used?

BearishSun:
Thanks for the explanation. This was merged in 6289fed0041614a7004f5efd6b8f72607b7bcfbd.

I've moved your two methods to `Time`, as well as most of the fixed update timing calculations as well.

BearishSun:
I think I see your point though, multiple iterations will only increment it once, which I guess is not adequate for your purposes?

ivalduan:
Yes, exactly. The point of having access to that value is to have the correct value on each fixedUpdate.

BearishSun:
Makes sense, I'll rethink that bit and let you know when its updated.

BearishSun:
I believe e4e4554a206c5ed3e40fb7d783d58807d0bc9dea should yield the same behavior as before.

PR #164, opened 2018-06-28
Bugfix: glMemoryBarrier() after draw/compute calls

w-flo:
Here's the first PR I promised. I hope I didn't miss any required glMemoryBarrier() spots.

### DIFF
--- a/Source/Plugins/bsfGLRenderAPI/BsGLRenderAPI.cpp
+++ b/Source/Plugins/bsfGLRenderAPI/BsGLRenderAPI.cpp
@@ -2155,7 +2155,9 @@ namespace bs { namespace ct
 		if(!mDrawCallInProgress)
 			return;
 
+#if BS_OPENGL_4_2 || BS_OPENGLES_3_1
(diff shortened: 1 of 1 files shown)


w-flo:
So the clang build fails:  error: use of undeclared identifier 'GL_ALL_BARRIER_BITS'

I'm not sure how to fix this and have no Apple hardware for testing, so I don't want to keep pushing fix attempts until clang likes it. :-) 

Maybe add
#ifdef GL_ARB_shader_image_load_store 
?

w-flo:
I have a #ifdef GL_ARB_shader_image_load_store check ready to be pushed, but since I'm not sure if it's the right way to do the check (is a runtime check required? Should it check for either the extension, or OpenGL >= 4.2, or OpenGL ES >= 3.1, since all three variants support memory barriers?) I'm waiting for comments about this.

BearishSun:
Yep I believe you figured it out, adding `#if BS_OPENGL_4_2 || BS_OPENGLES_3_1` should be the right call. Other than that, everything looks good.

BearishSun:
Merged, thanks!

Issue #168, opened 2018-07-03
Problem with basic shadow setups with radial or spot light


ivalduan:
I attach a scene as simplified as possible to reproduce.
https://drive.google.com/open?id=12ylmbXk5nnsNKbXgBfZJULdqU0dHHk-E

The only light type that casts shadows correctly in the scene is directional.
Tested with Windows and DirectX11/Vulkan APIs.

Also sometimes a spot light appears and disappears from view moving inside a scene, so there seems to be a problem with the visibility of this type of light, I will try to reproduce in another example.

ivalduan:
It seems that (opt.intersects(bounds)) fails inside ShadowRenderQueue::execute even if a cube or a sphere are inside the light volume, the plane casts shadows correctly on the cube.
bounds is good so it has to be opt frustum planes.

BearishSun:
Thanks for the report. I'll take a closer look later, but since you seem to be investigating, take a look at PR #165 . It changed the order of frustum planes in a convex volume which could have possibly affected the bounds tests. It's a longshot but its the only thing in that system that has been recently changed. I'll dig deeper when I get a chance.

w-flo:
Sorry about that, I should have done a better job checking this when suggesting to change the plane ordering. The collection itself is only used for iteration, but of course that changes the insertion order for other collections, even including CCamera::getWorldFrustum(), and I failed to take that into account. Maybe issue #169 is related? Should PR #165 be reverted / changed to restore the old plane ordering, in order to prevent any further surprises? Scanning the various files I can't find any other obvious dependency on plane order, but it's a risk.

BearishSun:
When you made that PR I vaguely remembered there was a plane ordering bit of code within the shadow rendering code that could be affected, but I couldn't find it either. No need to revert the PR as I'm pretty sure the ordering is not relevant elsewhere. (But if we had the ordering in the first place this issue wouldn't have happened as I would have used the enum instead of referencing the far plane in this way :) ).

Issue #169 doesn't appear to be related, it seems to be just an issue with spot light bound calculation.

PR #170, opened 2018-07-03
Bugfix: Problem with basic shadow setups with radial or spot light (#168)

ivalduan:
The problem with radial lights was related with the ordering of the planes as you guessed.

The problem with spot lights is a problem with matrix transforms, there is some computations that I am not grasping completely, but the bug seems to be fixed by getting the correct worldMatrix, seems that previous approach was as the transpose of view, but view is not orthogonal so this is not equivalent to the inverse.


BearishSun:
Thanks for taking the time to track it down, as well as providing reproducible test case!

I wasn't able to merge your PR directly because I guess it diverged too much from upstream. git just wasn't able to figure it out - rebase just plain skipped one of the commits and merge introduced changes that were already part of the branch for about 40 files.

So in the end I just made the changes manually. ea7332a65ab13eca3225755454bb840e8a2075ee

ivalduan:
Hmm weird, I merged the current state of upstream to my fork before PR.

Just out of curiosity, what is the recommended way of handling small PRs with github?
I thought having a sync fork was the only way but maybe it is better to open a new branch per each PR from the most recent upstream/master commit and then closing it.

BearishSun:
Not really sure what confused git, when I did a merge it did a fast-forward and if everything seemed fine, but when I looked one of the commits it contained a bunch of changes from previous commits that were already part of that branch.

I think the cleanest way is to open up a new branch for each PR as you suggested.

jonesmz:
What I always do is
1) Make sure my "master" branch is always identical to upstream.
2) For any feature branches that I'm creating, if I update my master branch, i "git rebase" my feature branch so that it's always adding my commits on top of what's in the master branch, and the git log shows a direct linear relationship instead of a fork.

It's more work, but I've had github's PR interface screw things up frequently, and git itself has trouble with merging frequently enough, that I think it's about break even as far as paying the effort ahead of time or after the fact.

PR #177, opened 2018-07-11
Adds 'ManagedStackAlloc' and 'ManagedStackNew' types with associated functions. Resolves #134

### LINKED ISSUE #134: Implement RAII way to free stack allocations
Basically a reminder to myself to make a pull request for this.

jonesmz:
Tested replacing a small handful of bs_stack_alloc and bs_stack_new calls with the new functions. Everything compiles.

I did not test this at runtime. Just compile time.

Documentation might not be perfect, sorry. Not super knowledgeable about doxygen.

Things that could be improved

Merged the two new structs into one, and use a constructor deduction guide to bind the address of a free function to a template parameter (so you don't need to store a pointer to function).

Remove the move constructors completely, when you switch to c++17. C++17 makes copy-elision mandatory, thus no longer performing a check for the existence of a move function even though it's never used (even in C++14, it wouldn't be used by any major compiler, because they're going to do RVO. They're just required to check that a move constructor exists, sadly.).

An alternative to doing it this way, is that you could have a StackAllocator, that does the cleanup.
I felt that the extra overhead of tracking types and sizes to avoid destructor issues wasn't worth it, maybe you can think of a way to not have to track that kind of information (Or track it without allocating).

BTW.

template<class T, class... Args>
T* bs_stack_new(Args &&...args, UINT32 count = 0)

Makes me nervous.

I've spent many an hour fighting MSVC 2015 on the existence of a parameter AFTER a parameter pack. It simply does not allow that, and I'm pretty sure it's correct that this isn't allowed.

Are you sure this is valid?

Another aside: There are at least many dozen places in the bsf codebase where you're casting the result of bs_stack_alloc() to the data type desired, and doing manual multiplication of the sizeof() the data type.

I'd recommend switching these all to the template version. (You'll need to if you use this new managed_stack_alloc code anyway).

### DIFF
--- a/Source/Foundation/bsfUtility/Allocators/BsStackAlloc.h
+++ b/Source/Foundation/bsfUtility/Allocators/BsStackAlloc.h
@@ -24,16 +24,16 @@ namespace bs
 	 * Describes a memory stack of a certain block capacity. See MemStack for more information.
 	 *
 	 * @tparam	BlockCapacity Minimum size of a block. Larger blocks mean less memory allocations, but also potentially
-	 *						  more wasted memory. If an allocation requests more bytes than BlockCapacity, first largest 
+	 *						  more wasted memory. If an allocation requests more bytes than BlockCapacity, first largest
 	 *						  multiple is used instead.
 	 */
 	template <int BlockCapacity = 1024 * 1024>
 	class MemStackInternal
 	{
 	private:
-		/** 
-		 * A single block of memory of BlockCapacity size. A pointer to the first free address is stored, and a remaining 
-		 * size. 
+		/**
+		 * A single block of memory of BlockCapacity size. A pointer to the first free address is stored, and a remaining
+		 * size.
 		 */
 		class MemBlock
 		{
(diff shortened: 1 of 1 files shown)


BearishSun:
Thanks! Looks great at first glance. I'll take a closer look and merge when I get some more free time.

I'm not sure that parameter is valid, could be that it's never used so I never encountered any issues.

BearishSun:
I think I'm casting result of `bs_stack_alloc` because the type templated version doesn't work as intended and I never bothered to fix it. Since there is `bs_stack_alloc<Alloc>` version whenever you specify the type it assumes you are specifying the allocator. So the proper version will only be deduced if you specify both a type and an allocator, which is even more hassle than casting.

BearishSun:
I noticed `ManagedStackAlloc` and `ManagedStackNew` are in the `impl` namespace but they are returned by the allocator methods. `impl` to me implies something that shouldn't be used by user code.

BearishSun:
Could we merge the two types into one and make a branch in the constructor based on `mCount`? If 0 assume `free` needs to be called. Yeah it's an extra branch + extra storage for `mCount` when it might not be needed, but since everything is in the header the compiler might be smart enough to optimize that out when not needed anyway (count equal zero vs. non-zero should always be known at compile time).

If it ends up being just one class, perhaps call it `StackMemory`.

jonesmz:
How's this version?

Also, I don't see any bs_stack_** functions that take an Alloc template parameter?

In a different file?

BearishSun:
Never mind about `bs_stack_**` with Alloc. I saw it pop up today in intellisense but I must have accidentally been calling `bs_alloc`.

I think the comparison in destructor should be `>= 1`, and the default `count` parameter for `new` methods `= 1`.

Other than that looks great, thanks for making the changes!

PR #181, opened 2018-07-17
Basic fixes for 3 kinds of compiler warnings

lysannschlegel:
Just some simple fixes for compiler warnings I came across (using clang on Xcode 9.2):

- [Suggest braces around initialization of subobject](https://github.com/GameFoundry/bsf/commit/6ba17be300ab0ebe81f5350019b2cb5c53aa7b7e)
- [Nested namespace definition is a C++1z extension; define each namespace separately](https://github.com/GameFoundry/bsf/commit/c23d34486b55d6e8cd759bfe9fff657fe4a8043d)
- [Comparison of integers of different signs: 'INT32' (aka 'int') and 'UINT32' (aka 'unsigned int')](https://github.com/GameFoundry/bsf/commit/2b8f6718b652b434c381c2112c528b3ed45ad992)
    - Not entirely sure about this exact fix. Might just want to change the type of `i` to `UINT32`, but that might have implications about conversions to `int` for the offset values. Don't know if we care about that at all?

### DIFF
--- a/Source/Foundation/bsfEngine/Utility/BsShapeMeshes3D.cpp
+++ b/Source/Foundation/bsfEngine/Utility/BsShapeMeshes3D.cpp
@@ -1202,7 +1202,7 @@ namespace bs
 			{
 				int offsetA = i == 0 ? numArcVertices - 1 : i - 1;
 				int offsetB = i;
-				int offsetC = i == numArcVertices ? 1 : i + 1;
+				int offsetC = i == (INT32)numArcVertices ? 1 : i + 1;
 
 				Vector3* a = (Vector3*)(outVertices + (offsetA * vertexStride));
 				Vector3* b = (Vector3*)(outVertices + (offsetB * vertexStride));
--- a/Source/Foundation/bsfCore/Scene/BsSceneManager.h
+++ b/Source/Foundation/bsfCore/Scene/BsSceneManager.h
@@ -213,7 +213,7 @@ namespace bs
 		Vector<HComponent> mUninitializedComponents;
 
 		std::array<Vector<HComponent>*, 3> mComponentsPerState = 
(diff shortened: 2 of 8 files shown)


Source/Foundation/bsfEngine/Utility/BsShapeMeshes3D.cpp
```diff
@@ -1202,7 +1202,7 @@ namespace bs
 			{
 				int offsetA = i == 0 ? numArcVertices - 1 : i - 1;
 				int offsetB = i;
-				int offsetC = i == numArcVertices ? 1 : i + 1;
+				int offsetC = i == (INT32)numArcVertices ? 1 : i + 1;
```
lysannschlegel: Not entirely sure about this exact fix. Might just want to change the type of `i` to `UINT32`, but that might have implications about conversions to `int` for the offset values. Don't know if we care about that at all?

jonesmz:
My two cents are that the nested namespace should be left alone.

They're part of the C++17 language standard. The XCode platform just needs to catch up.

lysannschlegel:
> My two cents are that the nested namespace should be left alone.
>
> They're part of the C++17 language standard. The XCode platform just needs to catch up.

I have no strong feelings about this. Was just easy to fix and the same style is used in most places in the project. And it made sense to me since the project at this point aims at C++14, not 17.

jonesmz:
Sure, no malice assumed :-).

Just my own opinion. @BearishSun is going to do what he think's is best.

Source/Foundation/bsfEngine/Utility/BsShapeMeshes3D.cpp
BearishSun: I think changing to `UINT32` should be safe. I'll make the change before I merge.

PR #182, opened 2018-07-26
Add epsilon to point on plain side calculation

bmsq:
Adds epsilon value when calculating Plane.getSide(point).  This is useful when floating point precision makes it near impossible to correctly identify when a point is on the plain.   For example, binary space partition tree calculation.


BearishSun:
Looks good, thanks.

PR #184, opened 2018-08-04
Generate mPointerDelta and event.delta before event is fired

hachmeister:
The delta information in the event onPointerMoved was 0 for the x and y coordinates, so I moved the generation of the delta coordinates before the event is fired.


BearishSun:
Looks good, thanks. Merged.

PR #202, opened 2018-10-07
BsCCapsuleCollider calculate local rotation from normal on construction

DNKpp:
There is currently an inconsistency when creating a BsCCapsuleCollider and setting its normal property. The default value for the normal property is `bs::Vector3::VECTOR_Y` but the initial rotation of the base bs::CCollider use the Quaternation of `bs::Vector3::VECTOR_X`. That's the reason why the Collider will be orientated in the wrong direction.
Additionally the current code checks in the setNormal function if the new normal differs from the current one, thus we can't change the normal `bs::Vector3::VECTOR_Y` to bypass this issue. The only solution is to first set it to `bs::Vector3::VECTOR_X` or `bs::Vector3::VECTOR_Z` and then reset it to `bs::Vector3::VECTOR_Y`. This commit fixes this bug.

### DIFF
--- a/Source/Foundation/bsfCore/Components/BsCCapsuleCollider.cpp
+++ b/Source/Foundation/bsfCore/Components/BsCCapsuleCollider.cpp
@@ -7,15 +7,25 @@
 
 namespace bs
 {
+	Quaternion _calculateRotationFromNormal(Vector3 normal)
+	{
+		normal.normalize();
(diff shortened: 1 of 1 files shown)


DNKpp:
Yeah, my mistake. I corrected it ;)

BearishSun:
Great, thank you. Merged.

PR #205, opened 2018-10-08
Fix `throw` being used by simdpp's aligned_allocator

### LINKED ISSUE #204: Build fails with clang 7 : `-fno-exceptions` vs throw.
This is caused by bsf disabling exceptions, and simdpp usage of a throw.
```
../bsf/Source/Foundation/bsfUtility/ThirdParty/simdpp/core/aligned_allocator.h:84:13: error: cannot use 'throw' with exceptions disabled
            throw std::length_error("aligned_allocator<T,A>::allocate() - Integer overflow.");
```
so should we replace this throw with an abort() ?

nemerle:
closes #204


BearishSun:
Thanks, looks good.

PR #212, opened 2018-10-20
Build: Add Windows Resource files and extra config values

guestnone:
This commit adds description for compiled binaries on Windows platforms
for easier version identification. It also adds additional configuration
values that can be set in main CMakeLists.txt file:
  * BS_VERSION_PATCH for very minor bug-fixing versions of the framework.
  * BS_CURRENT_RELEASE_YEAR which is self explanatory.

Plus the BS_VERSION_STRING macro for providing the version of the framework as a string.


BearishSun:
Looks good, thank you.

PR #213, opened 2018-10-21
Vulkan to Metal - MacOSX - integration - WIP

remoe:
This compiles and links with VulkanSDK and activated Vulkan API on MacOSX. I've build it with XCode 10 and MacOSX 10.13.6

Requirements:
VulkanSDK >= 1.1.92.1

OLD:
MoltenVK 1.0.24 (because VK_FORMAT_B10G11R11_UFLOAT_PACK32 format support)
I've updated vk_mem_alloc.h to latest version. It is a partial Vulkan 1.1 port as well.


remoe:
After try to run, i see issues with validation and utility layers for Vulkan:

```
[***MoltenVK ERROR***] VK_ERROR_LAYER_NOT_PRESENT: Vulkan layer VK_LAYER_LUNARG_standard_validation is not supported.
[***MoltenVK ERROR***] VK_ERROR_EXTENSION_NOT_PRESENT: Vulkan extension VK_EXT_debug_report is not supported.
[mvk-info] MoltenVK version 1.0.25. Vulkan version 1.0.85.
	The following Vulkan extensions are supported:
		VK_KHR_dedicated_allocation v3
		VK_KHR_descriptor_update_template v1
		VK_KHR_get_memory_requirements2 v1
		VK_KHR_get_physical_device_properties2 v1
		VK_KHR_image_format_list v1
		VK_KHR_maintenance1 v2
		VK_KHR_maintenance2 v1
		VK_KHR_push_descriptor v2
		VK_KHR_sampler_mirror_clamp_to_edge v1
		VK_KHR_shader_draw_parameters v1
		VK_KHR_surface v25
		VK_KHR_swapchain v70
		VK_EXT_shader_viewport_index_layer v1
		VK_EXT_vertex_attribute_divisor v3
		VK_MVK_macos_surface v2
		VK_MVK_moltenvk v11
		VK_AMD_negative_viewport_height v1
```

For this we need the LunarSDK:

https://vulkan.lunarg.com/doc/sdk/latest/mac/getting_started.html 

?

Source/Plugins/bsfVulkanRenderAPI/BsVulkanCommandBuffer.cpp
```diff
@@ -822,6 +824,8 @@ namespace bs { namespace ct
 				Win32RenderWindow* window = static_cast<Win32RenderWindow*>(rt.get());
 #elif BS_PLATFORM == BS_PLATFORM_LINUX
 				LinuxRenderWindow* window = static_cast<LinuxRenderWindow*>(rt.get());
+#elif BS_PLATFORM == BS_PLATFORM_OSX
+                MacOSRenderWindow* window = static_cast<MacOSRenderWindow*>(rt.get());
```
BearishSun: Please use tabs instead of spaces.

Source/Plugins/bsfVulkanRenderAPI/MacOS/BsMacOSRenderWindow.mm
```diff
@@ -0,0 +1,597 @@
...
+			bs_delete(mWindow);
+			mWindow = nullptr;
+		}
+
+		RenderWindow::destroy();
+	}
+
+	SPtr<ct::CoreObject> MacOSRenderWindow::createCore() const
+	{
+		/*
```
BearishSun: We can get rid of this comment.

Source/Plugins/bsfVulkanRenderAPI/MacOS/BsMacOSRenderWindow.mm
```diff
@@ -0,0 +1,597 @@
...
+		*/
+        
+		ct::VulkanRenderAPI& rapi = static_cast<ct::VulkanRenderAPI&>(ct::RenderAPI::instance());
+
+		RENDER_WINDOW_DESC desc = mDesc;
+		SPtr<ct::CoreObject> coreObj = bs_shared_ptr_new<ct::MacOSRenderWindow>(desc, mWindowId, mWindow->_getWindowId(), rapi);
+		coreObj->_setThisPtr(coreObj);
+
+		return coreObj;
+
```
BearishSun: Also this newline

Source/Plugins/bsfVulkanRenderAPI/MacOS/BsMacOSRenderWindow.mm
```diff
@@ -0,0 +1,597 @@
...
+	}
+
+	SPtr<ct::MacOSRenderWindow> MacOSRenderWindow::getCore() const
+	{
+		return std::static_pointer_cast<ct::MacOSRenderWindow>(mCoreSpecific);
+	}
+    
+    void MacOSRenderWindow::_windowMovedOrResized()
+    {
+        THROW_IF_NOT_CORE_THREAD;
```
BearishSun: This seems to have a lot less code than the OpenGL version, any reason?

Source/Plugins/bsfVulkanRenderAPI/MacOS/BsMacOSRenderWindow.mm
```diff
@@ -0,0 +1,597 @@
...
+		}
+
+		void MacOSRenderWindow::swapBuffers(UINT32 syncMask)
+		{
+            THROW_IF_NOT_CORE_THREAD;
+            
+            if (mShowOnSwap)
+                setHidden(false);
+            
+            //LinuxPlatform::lockX();
```
BearishSun: Extra comment

Source/Plugins/bsfVulkanRenderAPI/MacOS/BsMacOSRenderWindow.mm
```diff
@@ -0,0 +1,597 @@
...
+                mSemaphoresTemp[numSemaphores] = mSwapChain->getBackBuffer().sync;
+                numSemaphores++;
+                
+                mSwapChain->notifyBackBufferWaitIssued();
+            }
+            
+            queue->present(mSwapChain.get(), mSemaphoresTemp, numSemaphores);
+            mRequiresNewBackBuffer = true;
+            
+            //LinuxPlatform::unlockX();
```
BearishSun: Extra comment

BearishSun:
Thanks for working on this. I added some comments to clean up some of the code.

Also could you rebase all your changes on top of the latest master?

I'm not really sure how validation layers work on macOS, but indeed on Windows & Linux they are part of the SDK.

Source/Plugins/bsfVulkanRenderAPI/MacOS/BsMacOSRenderWindow.mm
remoe: i redirected it to ct::BsMacOSRenderWindow

BearishSun: This method will get called on the main thread, while everything in the ct:: namespace executes on the core thread. It's not safe to call core thread methods directly, plus its method does different things like rebuilding the swap chain, while this method should just update some local values. Also `THROW_IF_NOT_CORE_THREAD` should trigger here, as this method shouldn't be called on the core thread unless I'm misremembering something.

remoe: ok, this is a very important thing, thanks ;)

remoe:
Ok, it's simple to get the layers and loader to work. One should use the LunarSDK and set the following parameters:

VK_ICD_FILENAMES= ... vulkan-osx-1.1.85/etc/vulkan/icd.d/MoltenVK_icd.json
VK_LAYER_PATH= ... vulkan-osx-1.1.85/etc/vulkan/explicit_layer.d

Then the Vulkan instance would be created :)

BearishSun:
MoltenVK doesn't support geometry shaders (nor does Metal), so you'll probably need exclude any code related to geometry shaders if compiling on macOS. You can probably follow those errors for locations where they might be used, and just to a search for anything `geometry`.

`bsf` shouldn't be using any geometry shaders on mac anyway, so this shouldn't break anything in the renderer.

remoe:
Thanks, i think i solved it.

remoe:
The "TBuiltInResource" on https://github.com/GameFoundry/bsf/blob/6b9d82d2b5724a6b9532ae69283f939dbaff89c5/Source/Plugins/bsfVulkanRenderAPI/Managers/BsVulkanGLSLProgramFactory.cpp#L19 doesn't match the ResourceLimits.h of Vulkan 1.1.85 gslang/include/ResourceLimits.h. 

How do you want to handle this?

remoe:
Oh, i see you have this dependency in our own code: Dependencies/glslang/include as well . I've build it before with macosx_deployment_target=10.9. Now i use "10.12". Didn't know why, but i don't have a compiler error on this ...

remoe:
it crashes now here: 

https://github.com/GameFoundry/bsf/blob/master/Source/Plugins/bsfVulkanRenderAPI/Managers/BsVulkanRenderWindowManager.cpp#L18

with:

```
EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)
libbsfVulkanRenderAPI.dylib`bs::VulkanRenderWindowManager::createImpl(this=0x00000001026290c0, desc=0x00007ffeefbfe800, windowId=0, parentWindow=nullptr) at BsVulkanRenderWindowManager.cpp:21
libbsf.dylib`bs::RenderWindowManager::create(this=0x00000001026290c0, desc=0x00007ffeefbfe800, parentWindow=nullptr) at BsRenderWindowManager.cpp:24
libbsf.dylib`bs::RenderWindow::create(desc=0x00007ffeefbfe800, parentWindow=nullptr) at BsRenderWindow.cpp:174
libbsf.dylib`bs::ct::RenderAPI::initialize(this=0x00000001043489a0, primaryWindowDesc=0x00000001043014f8) at BsRenderAPI.cpp:168
libbsf.dylib`bs::RenderAPIManager::initialize(this=0x00000001045224a0, pluginFilename=0x0000000104301478, primaryWindowDesc=0x00000001043014f8) at BsRenderAPIManager.cpp:46
libbsf.dylib`bs::CoreApplication::onStartUp(this=0x0000000104301460) at BsCoreApplication.cpp:154
libbsf.dylib`bs::Application::onStartUp(this=0x0000000104301460) at BsApplication.cpp:52
bsfHelloWorld`bs::MyApplication::onStartUp(this=0x0000000104301460) at main.cpp:72
bsfHelloWorld`void bs::Module<bs::CoreApplication>::startUp<bs::MyApplication, bs::START_UP_DESC&>(args=0x00007ffeefbff488) at BsModule.h:92
bsfHelloWorld`void bs::Application::startUp<bs::MyApplication>(videoMode=(mWidth = 1280, mHeight = 720, mRefreshRate = 60, mOutputIdx = 0, mIsCustom = true), title=0x00007ffeefbff608, fullscreen=false) at BsApplication.h:33
```

It's looks like a binary corruption. But i didn't know from where this should come from.

BearishSun:
We use glslang internally to translate GLSL to SPIR-V, and I believe MoltenVK uses it for the same purpose. Not sure if it's needed in MoltenVK since bsf has it, but even if it's not I'm not sure if there's a way to disable it or how to stop it conflicting with our version.

remoe:
@BearishSun on vkCreateShaderModule i have now:

A fatal error occurred and the program has to terminate!
  - Error: RenderingAPIException
  - Description: ERROR: [Validation] Code 0:  [ UNASSIGNED-CoreValidation-Shader-InconsistentSpirv ] Object: VK_NULL_HANDLE (Type = 0) | SPIR-V module not valid: MemoryBarrier: Vulkan specification requires Memory Semantics to have one of the following bits set: Acquire, Release, AcquireRelease or SequentiallyConsistent
  OpMemoryBarrier %uint_1 %uint_512

**Another help call to speed things up** :)

remoe:
the build failed on CI only because the CI failed :) 

`AppVeyor failed to assign job to 'gce' build environment after 3 attempt(s). The envronment is either unavailable or there are no free workers, however alternate build env`

BearishSun:
Those kind of errors tend to be caused by `glslang`, which generates the SPIR-V code that gets fed into Vulkan. In this case it seems its generating something the specified Vulkan version doesn't like. Usually I solve this by upgrading `glslang` to a more recent version. 

If that doesn't help then maybe there are some flags that need to be toggled when `glslang` is called (perhaps even something MoltenVK specific).

And finally, if none of that works, you can check on exactly which GPU program is it failing (its source code should be present on the `GpuProgram` object, which should be close to where the error is triggering). Then potentially tweak the program itself to avoid using the instruction thats being a problem. But ideally this isn't something you need to do.

remoe:
ok, thanks. I solved it with update to a newer glslang.

remoe:
Next Vulkan error from here:

```c++
VkResult result = vkCreateImageView(mOwner->getDevice().getLogical(), &mImageViewCI, gVulkanAllocator, &view);
```

  - Error: RenderingAPIException
  - Description: ERROR: [Validation] Code 0:  [ VUID-VkImageViewCreateInfo-image-01019 ] Object: 0x4be (Type = 10) | vkCreateImageView() format VK_FORMAT_B10G11R11_UFLOAT_PACK32 differs from image 1214 format VK_FORMAT_R8G8B8A8_UNORM.  Formats MUST be IDENTICAL unless VK_IMAGE_CREATE_MUTABLE_FORMAT BIT was set on image creation. The Vulkan spec states: If image was not created with the VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT flag, format must be identical to the format used to create image (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkImageViewCreateInfo-image-01019)

BearishSun:
It seems the image was created with `VK_FORMAT_R8G8B8A8_UNORM` format, but the image view is being created with `VK_FORMAT_B10G11R11_UFLOAT_PACK32` format. I can't tell you more without more information, but you can probably dig through the callstack when that happens and find the corresponding `VulkanImage`, and perhaps track down why is it being created with a different format?

remoe:
yes, sorry, here is it:
The main issue is that getClosestSupportedPixelFormat returns for the format PF_RG11B10F the PF_RGBA8 because vkGetPhysicalDeviceFormatProperties return false and the result comes from here:

https://github.com/GameFoundry/bsf/blob/master/Source/Plugins/bsfVulkanRenderAPI/BsVulkanUtility.cpp#L72

BearishSun:
Hmm so that format is not supported? I find that strange since I see the format in Metal specification, and pretty much all the hardware I'm aware of supports it. Could you run `vulkaninfo` (should be somewhere in the Vulkan SDK) and see if it prints out the format? Might be something else we're doing wrong.

If it's really not supported we'll probably want to use the 16-bit per-channel floating point format instead (and fix the view creation so it realizes when the exact format wasn't available).

remoe:
yes the format should be supported:

FORMAT_B10G11R11_UFLOAT_PACK32:
	linearTiling   FormatFeatureFlags:
		None
	optimalTiling  FormatFeatureFlags:
		None
	bufferFeatures FormatFeatureFlags:
		None

I need to find the reason why this not work ...

BearishSun:
That looks like it isn't supported. You should be getting some feature flags like `VK_FORMAT_FEATURE_SAMPLED_IMAGE_BIT`, etc.

Looking closer, it seems this format was removed from `MetalVK` in 1.0.16:
> Remove support for VK_FORMAT_B10G11R11_UFLOAT_PACK32 & VK_FORMAT_E5B9G9R9_UFLOAT_PACK32 since format components are reversed on Metal.

Full link: https://github.com/KhronosGroup/MoltenVK/blob/master/Docs/Whats_New.md

I think we'll just need to go with the 16-bit per channel floating point fallback approach.

remoe:
This is solved here:

https://github.com/KhronosGroup/MoltenVK/pull/304/commits/e7af5bed64e1e51bafb47a70887df6d45a27751e

The latest master of MoltenVK didn't have this issue.

remoe:
Now, vulkan cannot create the pipeline, because the shader compiler failed:

[***MoltenVK ERROR***] VK_ERROR_FORMAT_NOT_SUPPORTED: Unable to convert SPIR-V to MSL:
MSL conversion error: Cannot yet map non-aliased parameter to Metal resource!

[***MoltenVK ERROR***] VK_ERROR_INITIALIZATION_FAILED: Fragment shader function could not be compiled into pipeline. See previous error.
Assertion failed: (result == VK_SUCCESS), function createPipeline, filebsf.git/Source/Plugins/bsfVulkanRenderAPI/BsVulkanGpuPipelineState.cpp, line 482.

BearishSun:
It seems something in "IrradianceProjectSHMat.bsl" shader is confusing it, but I don't have an idea what exactly could that error be referring to. That shader uses pretty standard shading functionality, no fancy stuff. Perhaps some mismatch between `glslang` used in `bsf` and whatever tool they use for converting SPIR-V to MSL? I think it's possible to pass raw GLSL to MoltenVK (`bsf` currently converts it to SPIR-V first), which might be worth a try?

You can also try modifying that shader file, reducing it to barebones and see if any part of it in particular is confusing it.

Source/Plugins/bsfVulkanRenderAPI/MacOS/BsMacOSRenderWindow.mm
```diff
@@ -30,6 +30,83 @@
...
+			[view setWantsLayer:YES];
+		}
+		
+		VkMacOSSurfaceCreateInfoMVK surfaceCreateInfo;
+		surfaceCreateInfo.sType = VK_STRUCTURE_TYPE_MACOS_SURFACE_CREATE_INFO_MVK;
+		surfaceCreateInfo.pNext = nullptr;
+		surfaceCreateInfo.flags = 0;
+		surfaceCreateInfo.pView = (__bridge const void *)(mWindow->getView());
+		
+		ct::VulkanRenderAPI& rapi = static_cast<ct::VulkanRenderAPI&>(ct::RenderAPI::instance());
```
BearishSun: Hmm, I think this will work but it's very sketchy as it's relying on the core thread actually initializing Vulkan at this point. Which I think it should because there is a sync point elsewhere in code, but relying on some sync point elsewhere might not be the safest approach. How come this needs to be on the main thread and can be it moved to core thread?

remoe: I know, but MoltenVK didn't allow to call vkCreateMacOSSurfaceMVK on non main thread.

remoe:
I checked if the shader file is loaded and i checked in VulkanGLSLProgramFactory::compileBytecode if the shader code compiles and links. I don't see any error. "GlslangToSpv" does not print any error.

Here are the shaders which goes through "VulkanGLSLProgramFactory::compileBytecode". Sorry not formatted:

 __l = (__cap_ = 1697, __size_ = 1695, __data_ = "// GLSL Compute Shader "csmain"\n// Generated by XShaderCompiler\n// 11/04/2018 10:34:18\n\n#version 450\n\nlayout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;\n\nstruct SHVector\n{\n    float v[25];\n};\n\nstruct SHVectorRGB\n{\n    SHVector R;\n    SHVector G;\n    SHVector B;\n};\n\nvoid SHZero(inout SHVector v)\n{\n    for (int i = 0; i < 5 * 5; ++i)\n        v.v[i] = 0.0f;\n}\n\nvoid SHAdd(inout SHVector lhs, SHVector rhs)\n{\n    for (int i = 0; i < 5 * 5; ++i)\n        lhs.v[i] += rhs.v[i];\n}\n\nvoid SHMultiply(inout SHVector lhs, float rhs)\n{\n    for (int i = 0; i < 5 * 5; ++i)\n        lhs.v[i] *= rhs;\n}\n\nstruct SHCoeffsAndWeight\n{\n    SHVectorRGB coeffs;\n    float       weight;\n};\n\nlayout(std430, binding = 0) buffer gInput\n{\n    SHCoeffsAndWeight xst_gInput[];\n};\n\nlayout(binding = 1) writeonly uniform image2D gOutput;\n\nlayout(std140, binding = 2) uniform Params\n{\n    uvec2 gOutputIdx;\n    uint  gNumEntries;\n};\n\nvoid main()\n{\n    SHVectorRGB coeffs;\n    float weight = 0.0f;\n    SHZero(coeffs.R);\n    SHZero(coeffs.G);\n")

 __l = (__cap_ = 449, __size_ = 434, __data_ = "// GLSL Vertex Shader "vsmain"\n// Generated by XShaderCompiler\n// 11/04/2018 10:34:18\n\n#version 450\n\nout gl_PerVertex\n{\n    vec4 gl_Position;\n};\n\nlayout(location = 0) in vec2 POSITION0;\nlayout(location = 1) in vec2 TEXCOORD0;\n\nlayout(location = 0) out vec2 bs_TEXCOORD0;\nlayout(location = 1) out vec2 bs_TEXCOORD1;\n\nvoid main()\n{\n    gl_Position = vec4(POSITION0, 0, 1);\n    bs_TEXCOORD0 = TEXCOORD0;\n    bs_TEXCOORD1 = POSITION0;\n}\n\n")

 __l = (__cap_ = 3857, __size_ = 3846, __data_ = "// GLSL Fragment Shader "fsmain"\n// Generated by XShaderCompiler\n// 11/04/2018 10:34:18\n\n#version 450\n\nlayout(location = 0) in vec2 bs_TEXCOORD0;\nlayout(location = 1) in vec2 bs_TEXCOORD1;\n\nlayout(location = 0) out vec4 SV_Target0;\n\nvec3 getDirFromCubeFace(uint cubeFace, vec2 uv)\n{\n    vec3 dir;\n    if (cubeFace == 0u)\n        dir = vec3(1.0f, -uv.y, -uv.x);\n    else if (cubeFace == 1u)\n        dir = vec3(-1.0f, -uv.y, uv.x);\n    else if (cubeFace == 2u)\n        dir = vec3(uv.x, 1.0f, uv.y);\n    else if (cubeFace == 3u)\n        dir = vec3(uv.x, -1.0f, -uv.y);\n    else if (cubeFace == 4u)\n        dir = vec3(uv.x, -uv.y, 1.0f);\n    else\n        dir = vec3(-uv.x, -uv.y, -1.0f);\n    return dir;\n}\n\nstruct SHVector\n{\n    float v[25];\n};\n\nstruct SHVectorRGB\n{\n    SHVector R;\n    SHVector G;\n    SHVector B;\n};\n\nSHVectorRGB SHLoad(sampler2D xsr_input, ivec2 offset)\n{\n    SHVectorRGB xsr_output;\n    for (int i = 0; i < 5 * 5; ++i)\n    {\n        ivec3 xst_temp0 = ivec3(offset.x + i, offset.y, 0);\n        vec3 coeff = t")
```

**Comment by remoe** [comment]
I've build MoltenVK in debug mode. You find the dumps above in pastebin.

**Comment by remoe** [comment]
Awesome hint! Thanks! I've changed the code to:

from:

```glsl
		float4 fsmain(VStoFS input) : SV_Target0
		{
			float2 scaledUV = input.uv0 * 2.0f - 1.0f;
			float3 dir = getDirFromCubeFace(gCubeFace, scaledUV);
			dir = normalize(dir);
			
			SHVector shBasis = SHBasis(dir);
							
			SHVectorRGB coeffs = SHLoad(gSHCoeffs, int2(0, 0));
```

to:

```glsl
		float4 fsmain(VStoFS input) : SV_Target0
		{
			float2 scaledUV = input.uv0 * 2.0f - 1.0f;
			float3 dir = getDirFromCubeFace(gCubeFace, scaledUV);
			dir = normalize(dir);
			
			SHVector shBasis = SHBasis(dir);

			int2 offset = int2(0, 0);				
			SHVectorRGB coeffs;

			#define SH_NUM_COEFFS SH_ORDER * SH_ORDER
			for(int i = 0; i < SH_NUM_COEFFS; ++i)
			{
				float3 coeff = gSHCoeffs.Load(int3(offset.x + i, offset.y, 0)).rgb;
			
				coeffs.R.v[i] = coeff.r;
				coeffs.G.v[i] = coeff.g;
				coeffs.B.v[i] = coeff.b;
			}
```

Now the sample work! A Cocoa window pops up. I see a rotating textured cube ! 🎉  📈 

@BearishSun , can you please do this bug report, because my bad english 😓 
Should i add this hack, or what we do to solve this in this PR?

BearishSun:
Awesome, nice work. I'll submit a bug report when I get a chance, if I don't forget.

For now we can solve it using the current hack, but keep the existing code commented out, and add a comment stating that the replacement code is a hack due to a bug in MoltenVK.

remoe:
FYI: I'm generally a windows developer :) I don't know much about MacOS and especially Objective-C, but i hope i can finish it anyway :)

remoe:
i don't understand why i have this error:

Cannot begin a sample because no frame is active.
		 in void bs::ProfilerGPU::beginSample(bs::ProfilerString) bsf.git/Source/Foundation/bsfCore/Profiling/BsProfilerGPU.cpp:68]

?

BearishSun:
All the `ProfilerGPU::beginSample` calls needs to be between `ProfilerGPU::beginFrame` and `ProfilerGPU::endFrame`. These calls are done automatically by the renderer, but if your code is getting called from somewhere else than the renderer this might happen. I can tell you more with a call stack of that method.

remoe:
```
libbsf.dylib`bs::ProfilerGPU::beginSample(this=0x000000010c848f10, name=bs::ProfilerString @ 0x0000700006603060) at BsProfilerGPU.cpp:69
libbsf.dylib`bs::ProfileGPUBlock::ProfileGPUBlock(this=0x0000700006603318, name=bs::ProfilerString @ 0x0000700006603298) at BsProfilerGPU.h:201
libbsf.dylib`bs::ct::RendererMaterialProfileBlock::RendererMaterialProfileBlock(this=0x0000700006603318, metaData=0x0000000100013760) at BsRendererMaterial.h:73
libbsf.dylib`bs::ct::RendererMaterialProfileBlock::RendererMaterialProfileBlock(this=0x0000700006603318, metaData=0x0000000100013760) at BsRendererMaterial.h:74
libbsf.dylib`bs::ct::BlitMat::execute(this=0x000000051b903c50, source=std::__1::shared_ptr<bs::ct::Texture>::element_type @ 0x00000005176fac30 strong=2 weak=2, area=0x0000700006603370, flipUV=false) at BsRendererUtility.cpp:373
libbsf.dylib`bs::ct::RendererUtility::blit(this=0x000000052383f100, texture=std::__1::shared_ptr<bs::ct::Texture>::element_type @ 0x00000005176fac30 strong=2 weak=2, area=0x0000000101279024, flipUV=false, isDepth=false) at BsRendererUtility.cpp:284
bsfHelloWorld`bs::ct::render() at main.cpp:364
bsfHelloWorld`void std::__1::__invoke_void_return_wrapper<void>::__call<void (*&)()>(void (*&&&)()) [inlined] decltype(__f=(0x000000051dd074b8))()>(fp)(std::__1::forward<>(fp0))) std::__1::__invoke<void (*&)()>(void (*&&&)()) at type_traits:4428
bsfHelloWorld`void std::__1::__invoke_void_return_wrapper<void>::__call<void (__args=(0x000000051dd074b8))()>(void (*&&&)()) at __functional_base:349
bsfHelloWorld`std::__1::__function::__func<void (*)(), std::__1::allocator<void (*)()>, void ()>::operator(this=0x000000051dd074b0)() at functional:1562
libbsf.dylib`std::__1::function<void ()>::operator(this=0x000000051dd074b0)() const at functional:1913
libbsf.dylib`bs::CommandQueueBase::playbackWithNotify(this=0x0000000516b0dcb0, commands=0x0000000520524f50 size=5, notifyCallback=0x0000000000000000)>) at BsCommandQueue.cpp:124
libbsf.dylib`bs::CommandQueueBase::playback(this=0x0000000516b0dcb0, commands=0x0000000520524f50 size=5) at BsCommandQueue.cpp:140
libbsf.dylib`void std::__1::__invoke_void_return_wrapper<void>::__call<std::__1::__bind<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>&>(std::__1::__bind<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>&&&) [inlined] decltype(__f=0x000000051787ac48, __a0=0x000000051787ac58, __args=size=0)).*fp(std::__1::forward<std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>(fp1))) std::__1::__invoke<void (bs::CommandQueueBase::*&)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&, void>(void (bs::CommandQueueBase::*&&&)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&&&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&&&) at type_traits:4369
libbsf.dylib`void std::__1::__invoke_void_return_wrapper<void>::__call<std::__1::__bind<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>&>(std::__1::__bind<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>&&&) [inlined] std::__1::__bind_return<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), std::__1::tuple<bs::CommandQueueBase*, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*>, std::__1::tuple<>, __is_valid_bind_return<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), std::__1::tuple<bs::CommandQueueBase*, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*>, std::__1::tuple<> >::value>::type std::__1::__apply_functor<void (__f=0x000000051787ac48, __bound_args=size=2, __args=size=0)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), std::__1::tuple<bs::CommandQueueBase*, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*>, 0ul, 1ul, std::__1::tuple<> >(void (bs::CommandQueueBase::*&)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), std::__1::tuple<bs::CommandQueueBase*, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*>&, std::__1::__tuple_indices<0ul, 1ul>, std::__1::tuple<>&&) at functional:2216
libbsf.dylib`void std::__1::__invoke_void_return_wrapper<void>::__call<std::__1::__bind<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>&>(std::__1::__bind<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>&&&) [inlined] std::__1::__bind_return<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), std::__1::tuple<bs::CommandQueueBase*, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*>, std::__1::tuple<>, __is_valid_bind_return<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), std::__1::tuple<bs::CommandQueueBase*, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*>, std::__1::tuple<> >::value>::type std::__1::__bind<void (this=0x000000051787ac48)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>::operator()<>() at functional:2249
libbsf.dylib`void std::__1::__invoke_void_return_wrapper<void>::__call<std::__1::__bind<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>&>(std::__1::__bind<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>&&&) [inlined] decltype(__f=0x000000051787ac48)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>&>(fp)(std::__1::forward<>(fp0))) std::__1::__invoke<std::__1::__bind<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>&>(std::__1::__bind<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>&&&) at type_traits:4428
libbsf.dylib`void std::__1::__invoke_void_return_wrapper<void>::__call<std::__1::__bind<void (__args=0x000000051787ac48)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>&>(std::__1::__bind<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>&&&) at __functional_base:349
libbsf.dylib`std::__1::__function::__func<std::__1::__bind<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&>, std::__1::allocator<std::__1::__bind<void (bs::CommandQueueBase::*)(std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*), bs::CommandQueueBase*&, std::__1::queue<bs::QueuedCommand, std::__1::deque<bs::QueuedCommand, bs::StdAlloc<bs::QueuedCommand, bs::GenAlloc> > >*&> >, void ()>::operator(this=0x000000051787ac40)() at functional:1562
libbsf.dylib`std::__1::function<void ()>::operator(this=0x000000051787ac40)() const at functional:1913
libbsf.dylib`bs::CommandQueueBase::playbackWithNotify(this=0x0000000102715720, commands=0x0000000102719ad0 size=0, notifyCallback=0x00007000066049c0)>) at BsCommandQueue.cpp:124```

**Comment by BearishSun** [comment]
I think that warning might be normal for that example, I don't have the time to check at the moment but it might be happening with other platforms/render backends.

**Comment by remoe** [comment]
Ok, no problem. I've tried SkeletalAnimation sample. This throws also an error:

  - Error: InvalidParametersException
  - Description: Wanted range is out of buffer bounds. Available range: 0 .. 28. Wanted range: 16 .. 32.
  - In function: void bs::ct::GpuParamBlockBuffer::write(bs::UINT32, const void *, bs::UINT32)
  - In file: bsf.git/Source/Foundation/bsfCore/RenderAPI/BsGpuParamBlockBuffer.cpp:127

**Comment by remoe** [comment]
About ProfilerGPU::beginSample: I've tested on Windows x64. It is the same behavior.

**Comment by remoe** [comment]
About SkeletalAnimation: I have the same issue with Windows x64! I use LunarSDK 1.1.85.0.

**Comment by BearishSun** [comment]
It's possible they both have the same cause. I haven't had a chance to try with newer Vulkan SDK myself. My best guess that the rules within `VulkanRenderAPI::generateParamBlockDesc` need to be updated to match whatever changed. When I wrote the method I tried to follow OpenGL `std140` layout rules (as described in 7.6.2.2 of [OpenGL specification](https://www.khronos.org/registry/OpenGL/specs/gl/glspec45.core.pdf)), but I might have gotten some edge case wrong.

**Comment by remoe** [comment]
I tried to find any 28 size block :) After many hours of debugging, i have some question: Is this correct:

https://github.com/GameFoundry/bsf/blob/6b9d82d2b5724a6b9532ae69283f939dbaff89c5/Source/Plugins/bsfVulkanRenderAPI/BsVulkanUtility.cpp#L758

? ( / 4)
I'm not sure with "CreateTonemapLUTParamDef". The size of this is 28 as well.

And this https://github.com/GameFoundry/bsf/blob/6b9d82d2b5724a6b9532ae69283f939dbaff89c5/Source/Plugins/bsfVulkanRenderAPI/Managers/BsVulkanGLSLProgramFactory.cpp#L545

does return 28 as well.

**Comment by BearishSun** [comment]
https://github.com/GameFoundry/bsf/blob/6b9d82d2b5724a6b9532ae69283f939dbaff89c5/Source/Plugins/bsfVulkanRenderAPI/Managers/BsVulkanGLSLProgramFactory.cpp#L545

Yep, I think this is the root cause of the issue. I think the solution might be just to round that size to a multiple of 16 bytes, always. 

In case that doesn't work, or breaks something else, you can add padding to the constant buffer in question, which is `MaterialParams` of `Diffuse.bsl` shader (judging by your call-stack) - adding an extra float to that buffer will pad it so it's always 32 bytes in size, and nothing should complain. Other shaders with similar buffer structures might need to be padded as well.

-----
Nice work getting surface creation to the core thread.

I'm aware of that swap chain issue, it also happens on Linux. I haven't gotten around to fixing it, but I think the swap chain needs to be rebuilt when `VK_ERROR_OUT_OF_DATE_KHR` is returned (it's an error message that can apparently be returned during normal operation). Although I remember I had some concerns regarding whether or not its safe to rebuild the swap chain at that stage, but I don't remember exactly why anymore.

**Comment by BearishSun** [comment]
Feel free to leave the resize issue to me, since I have it on my agenda anyway. But I certainly wont complain if you fix it :)

**Comment by remoe** [comment]
Sorry to say, but i think you need to rewrite/consolidate most of the window resize code, because BSF has a lot of resize "event" handlers:

RenderWindow::_windowMovedOrResized
bs::RenderWindowManager::instance().notifyMovedOrResized(this);
mutable Event<void()> onResized;

And one need to pay attention that some code need to run on main thread and other need to run on core thread.

You have solved most SwapChain issues right, but you need to refactor :) When you call something like "_windowMovedOrResized" here:

https://github.com/GameFoundry/bsf/blob/f133f3785da3058bff7d293a30276c895d3e5cf6/Source/Foundation/bsfCore/RenderAPI/BsRenderWindow.cpp#L199

One can handle the SwapChain rebuild before it will used in an invalid state?.

**Comment by BearishSun** [comment]
Feel free to offer ideas on how you'd prefer it to be structured. Note the code needs to interact with OS windowing code on multiple platforms, and they all handle things differently. 

`RenderWindow::_notifyWindowEvent` is the entry point for all window events, including move/resize. It will get called on core thread on Windows & Linux, main thread on macOS. It will call `RenderWindow::_windowMovedOrResized` override which does `RenderWindow` render backend and platform independant resize/move operations.

It also notifies the manager through `RenderWindowManager::notifyMovedOrResized`, which triggers the `RenderWindow::onResized` event (always on the main thread), which is the only handler intended for the user. `RenderWindow::resize()` just lets the user programmatically resize the window.

**Comment by BearishSun** [comment]
I believe `VK_ERROR_OUT_OF_DATE_KHR` happens before the OS notifies us the window got resized (i.e. before `RenderWindow::_windowMovedOrResized`). Leave it to me, I'll look into it more closely when I have time.

**Comment by remoe** [comment]
Ok, yeah it involve a lot of components in the framework. I know too little about it ;)

**Comment by remoe** [comment]
About alignment: I've changed the code to (Win x64):

```c++
int size = ((program->getUniformBlockSize(i) + 15) / 16) * 16;
```

Now, i have a different error:

  - Error: RenderingAPIException
  - Description: ERROR: [Validation] Code 0:  [ UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotUpdated ] Object: 0x74c (Type = 23) | Descriptor set 0x74c bound as set #0 encountered the following validation error at vkCmdDispatch() time: Descriptor in binding #0 at global descriptor index 0 requires an image view of type VK_IMAGE_VIEW_TYPE_2D_ARRAY but got VK_IMAGE_VIEW_TYPE_2D.

  - In function: unsigned int __cdecl bs::ct::debugMsgCallback(unsigned int,enum VkDebugReportObjectTypeEXT,unsigned __int64,unsigned __int64,int,const char *,const char *,void *)
  - In file: bsf.git\source\plugins\bsfvulkanrenderapi\bsvulkanrenderapi.cpp:77

BearishSun:
I'd have to see the call stack to narrow it down, but image view for some texture is not getting created correctly, in `VulkanImage::createView`. The calling code is likely trying to bind a single slice of a texture array, and the view is then being created as a normal texture, but the shader expects an array. We might need to force the view to always be a texture array if the image is a texture array.

This might be broken on other platforms as well as I changed some code related to this recently.

remoe:
here is the callstack:

```
 	VkLayer_core_validation.dll!log_msg(const _debug_report_data * debug_data, unsigned int msg_flags, VkDebugReportObjectTypeEXT object_type, unsigned __int64 src_object, std::basic_string<char,std::char_traits<char>,std::allocator<char> > vuid_text, const char * format, ...) Line 873	C++
 	VkLayer_core_validation.dll!core_validation::ValidateCmdBufDrawState(core_validation::layer_data * dev_data, GLOBAL_CB_NODE * cb_node, CMD_TYPE cmd_type, const bool indexed, const VkPipelineBindPoint bind_point, const char * function, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & pipe_err_code, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & state_err_code) Line 1248	C++
 	VkLayer_core_validation.dll!core_validation::ValidateCmdDrawType(core_validation::layer_data * dev_data, VkCommandBuffer_T * cmd_buffer, bool indexed, VkPipelineBindPoint bind_point, CMD_TYPE cmd_type, GLOBAL_CB_NODE * * cb_state, const char * caller, unsigned int queue_flags, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & queue_flag_code, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & renderpass_msg_code, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & pipebound_msg_code, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & dynamic_state_msg_code) Line 7481	C++
 	VkLayer_core_validation.dll!core_validation::PreCallValidateCmdDispatch(core_validation::layer_data * dev_data, VkCommandBuffer_T * cmd_buffer, bool indexed, VkPipelineBindPoint bind_point, GLOBAL_CB_NODE * * cb_state, const char * caller) Line 7660	C++
 	VkLayer_core_validation.dll!core_validation::CmdDispatch(VkCommandBuffer_T * commandBuffer, unsigned int x, unsigned int y, unsigned int z) Line 7674	C++
 	VkLayer_object_tracker.dll!object_tracker::CmdDispatch(VkCommandBuffer_T * commandBuffer, unsigned int groupCountX, unsigned int groupCountY, unsigned int groupCountZ) Line 1895	C++
 	VkLayer_parameter_validation.dll!parameter_validation::vkCmdDispatch(VkCommandBuffer_T * commandBuffer, unsigned int groupCountX, unsigned int groupCountY, unsigned int groupCountZ) Line 5645	C++
 	VkLayer_threading.dll!threading::CmdDispatch(VkCommandBuffer_T * commandBuffer, unsigned int groupCountX, unsigned int groupCountY, unsigned int groupCountZ) Line 2371	C++
 	bsfVulkanRenderAPI.dll!bs::ct::VulkanCmdBuffer::dispatch(unsigned int numGroupsX, unsigned int numGroupsY, unsigned int numGroupsZ) Line 1607	C++
 	bsfVulkanRenderAPI.dll!bs::ct::VulkanRenderAPI::dispatchCompute(unsigned int numGroupsX, unsigned int numGroupsY, unsigned int numGroupsZ, const std::shared_ptr<bs::ct::CommandBuffer> & commandBuffer) Line 466	C++
 	bsfRenderBeast.dll!bs::ct::ClearLoadStoreMat::execute(const std::shared_ptr<bs::ct::Texture> & target, const bs::TextureSurface & surface) Line 199	C++
 	bsfRenderBeast.dll!bs::ct::RCNodeLightAccumulation::render(const bs::ct::RenderCompositorNodeInputs & inputs) Line 730	C++
 	bsfRenderBeast.dll!bs::ct::RenderCompositor::execute(bs::ct::RenderCompositorNodeInputs & inputs) Line 198	C++
 	bsfRenderBeast.dll!bs::ct::RenderBeast::renderView(const bs::ct::RendererViewGroup & viewGroup, bs::ct::RendererView & view, const bs::ct::FrameInfo & frameInfo) Line 491	C++
 	bsfRenderBeast.dll!bs::ct::RenderBeast::renderViews(bs::ct::RendererViewGroup & viewGroup, const bs::ct::FrameInfo & frameInfo) Line 442	C++
 	bsfRenderBeast.dll!bs::ct::RenderBeast::renderAllCore(bs::ct::FrameTimings timings, bs::PerFrameData perFrameData) Line 405	C++
```

BearishSun:
It's the `Clear.bsl` shader. The issue is likely what I said above.

remoe:
Should I create a PR for the alignment issue?
What else do I have to do so that the PR can be merged?
When i find time i can try to solve the "createView" issue when you want.

BearishSun:
Feel free to make a PR or just add it to this one - either way is good for me. I can review & merge if you are not planning on making any more changes, just let me know. Feel free to investigate the createView issue as well.

remoe:
ok, great! Do you find time for this resize issue? It is one important thing which doesn't work current in this PR.

BearishSun:
Haven't had a chance to look at that yet, not sure when I will.

BearishSun:
Sorry, the shader is probably `ClearLoadStore.bsl` rather than `Clear.bsl`.

BearishSun:
The image view is correctly created in either case, it just doesn't match what the shader expects. I believe your fix prevents changing a view from array -> non-array, but that should be handled in the switch statement, not by permanently changing the image's view, since that could have side effects.

Some of those warnings could be normal.

I cannot give you any information about the error without more context.

remoe:
Ok, i can change the line:
https://github.com/GameFoundry/bsf/blob/master/Source/Plugins/bsfVulkanRenderAPI/BsVulkanTexture.cpp#L195

to

```c++
if (numFaces > 1 || mNumFaces > 1)
```

This is not nice, but it solve this case.

The TextureSurface:

```c++
TextureSurface surface;
surface.face = i;
surface.numFaces = 1; 
surface.mipLevel = 0;
surface.numMipLevels = 1;
```

is not sufficient to find that this need to be defined with VK_IMAGE_VIEW_TYPE_2D_ARRAY.

The callstack of the other error is:

```
ValidateCmdBufImageLayouts() - 0x140710784590752 File[buffer_validation.cpp:2953 (0)] Module[VkLayer_core_validation.dll]
core_validation::PreCallValidateQueueSubmit() - 0x140710783851095 File[core_validation.cpp:3114 (32)] Module[VkLayer_core_validation.dll]
core_validation::QueueSubmit() - 0x140710783855839 File[core_validation.cpp:3146 (22)] Module[VkLayer_core_validation.dll]
object_tracker::QueueSubmit() - 0x140710778317721 File[object_tracker.cpp:292 (0)] Module[VkLayer_object_tracker.dll]
parameter_validation::vkQueueSubmit() - 0x140710706921067 File[parameter_validation.cpp:2404 (21)] Module[VkLayer_parameter_validation.dll]
threading::QueueSubmit() - 0x140710702462098 File[thread_check.h:154 (0)] Module[VkLayer_threading.dll]
bs::ct::VulkanQueue::submitQueued() - 0x140710798376088 File[bsvulkanqueue.cpp:114 (48)] Module[bsfVulkanRenderAPI.dll]
bs::ct::VulkanCmdBuffer::submit() - 0x140710797848772 File[bsvulkancommandbuffer.cpp:699 (0)] Module[bsfVulkanRenderAPI.dll]
bs::ct::VulkanCommandBuffer::submit() - 0x140710797852212 File[bsvulkancommandbuffer.cpp:2433 (0)] Module[bsfVulkanRenderAPI.dll]
bs::ct::VulkanRenderAPI::submitCommandBuffer() - 0x140710797372287 File[bsvulkanrenderapi.cpp:552 (0)] Module[bsfVulkanRenderAPI.dll]
bs::ct::VulkanRenderAPI::swapBuffers() - 0x140710797372747 File[bsvulkanrenderapi.cpp:524 (65)] Module[bsfVulkanRenderAPI.dll]
bs::ct::RenderBeast::renderAllCore() - 0x140710697748963 File[bsrenderbeast.cpp:406 (88)] Module[bsfRenderBeast.dll]
std::_Invoker_pmf_pointer::_Call<void (__cdecl bs::ct::RenderBeast::*)(bs::ct::FrameTimings,bs::PerFrameData),bs::ct::RenderBeast * &,bs::ct::FrameTimings &,bs::PerFrameData &>() - 0x140710697418877 File[type_traits:16707566 (157)] Module[bsfRenderBeast.dll]
std::invoke<void (__cdecl bs::ct::RenderBeast::*&)(bs::ct::FrameTimings,bs::PerFrameData),bs::ct::RenderBeast * &,bs::ct::FrameTimings &,bs::PerFrameData &>() - 0x140710697500996 File[type_traits:16707566 (164)] Module[bsfRenderBeast.dll]
std::_Invoker_ret<std::_Unforced,0>::_Call<void (__cdecl bs::ct::RenderBeast::*&)(bs::ct::FrameTimings,bs::PerFrameData),bs::ct::RenderBeast * &,bs::ct::FrameTimings &,bs::PerFrameData &>() - 0x140710697417604 File[type_traits:15732480 (0)] Module[bsfRenderBeast.dll]
std::_Call_binder<std::_Unforced,0,1,2,void (__cdecl bs::ct::RenderBeast::*)(bs::ct::FrameTimings,bs::PerFrameData),std::tuple<bs::ct::RenderBeast *,bs::ct::FrameTimings,bs::PerFrameData>,std::tuple<> >() - 0x140710697419307 File[functional:1859 (0)] Module[bsfRenderBeast.dll]
std::_Binder<std::_Unforced,void (__cdecl bs::ct::RenderBeast::*)(bs::ct::FrameTimings,bs::PerFrameData),bs::ct::RenderBeast *,bs::ct::FrameTimings &,bs::PerFrameData &>::operator()<>() - 0x140710697415240 File[functional:1914 (136)] Module[bsfRenderBeast.dll]
std::_Invoker_functor::_Call<std::_Binder<std::_Unforced,void (__cdecl bs::ct::RenderBeast::*)(bs::ct::FrameTimings,bs::PerFrameData),bs::ct::RenderBeast *,bs::ct::FrameTimings &,bs::PerFrameData &> &>() - 0x140710697418254 File[type_traits:16707566 (62)] Module[bsfRenderBeast.dll]
std::invoke<std::_Binder<std::_Unforced,void (__cdecl bs::ct::RenderBeast::*)(bs::ct::FrameTimings,bs::PerFrameData),bs::ct::RenderBeast *,bs::ct::FrameTimings &,bs::PerFrameData &> &>() - 0x140710697501454 File[type_traits:16707566 (62)] Module[bsfRenderBeast.dll]
std::_Invoker_ret<void,1>::_Call<std::_Binder<std::_Unforced,void (__cdecl bs::ct::RenderBeast::*)(bs::ct::FrameTimings,bs::PerFrameData),bs::ct::RenderBeast *,bs::ct::FrameTimings &,bs::PerFrameData &> &>() - 0x140710697418350 File[type_traits:15732480 (0)] Module[bsfRenderBeast.dll]
std::_Func_impl_no_alloc<std::_Binder<std::_Unforced,void (__cdecl bs::ct::RenderBeast::*)(bs::ct::FrameTimings,bs::PerFrameData),bs::ct::RenderBeast *,bs::ct::FrameTimings &,bs::PerFrameData &>,void>::_Do_call() - 0x140710697624365 File[functional:15732480 (0)] Module[bsfRenderBeast.dll]
std::_Func_class<void>::operator()() - 0x140710093094976 File[functional:15732480 (0)] Module[bsf.dll]
```

BearishSun:
That's probably fine, API like DirectX doesn't support rebinding an array texture as a normal 2D texture anyway. The same check should be added for the 1D texture case.

I can't tell what's causing the other error. Obviously the image layout doesn't get set correctly but that could be due to a lot of things. If the image is bound for shader reads as the error implies, then the layout change should be initiated by `VulkanGpuParams::prepareForBind`, which in turns calls `VulkanCmdBuffer::registerResource` with the required layout, after which `VulkanCmdBuffer` takes care of sending the pipeline barrier that does the layout change. Something in that chain seems to be failing.

I suggest trying to track down exactly which texture is this happening for. Perhaps add a trace log where you print an VkImage ID + stack trace of every image that gets created (placed somewhere in `VulkanImage` probably), and then when the error triggers you can look up the image ID to see where the image originated and perhaps where it was first used.

After that you can ideally track down exactly which `VulkanGpuParams::prepareForBind` is binding that texture. After that you can do a line by line debug and drill into the code and see why it's not scheduling a layout change.

It's not an easy issue to track down sadly.

remoe:
Thanks for the detailed explanation! It is 

https://github.com/GameFoundry/bsf/blob/a034d72ab104e96a7e3296a9010368b4112bdb6a/Source/Plugins/bsfRenderBeast/BsRenderBeastIBLUtility.cpp#L386

But why ? :)

BearishSun:
Because you are using a newer Vulkan SDK, the last one I tested it properly on is 1.0.65.1.

My guess is there's a bug somewhere in the layout tracking code, within VulkanRenderAPI. It's probably in `VulkanCmdBuffer::registerResource(VulkanImage*, const VkImageSubresourceRange&, VkImageLayout, VkImageLayout, VulkanUseFlags, ResourceUsage)`. That method isn't the simplest one to follow - but the gist of it is that it tries to keep track of VkImageLayout for each subresource (face & mip level) of an image, and issue any layout changes if necessary.

remoe:
btw. The sample runs when the validation layer is disabled.

BearishSun:
When running outside of debug mode validation layers should not be active. I believe it's a good idea to always keep them active in debug mode.

Unfortunate about the MSL issue. I don't know enough about MSL to know if that's a limitation of the language itself, or just a fault in the conversion?

BearishSun:
Yeah that line doesn't seem to do anything, probably a leftover of some other code. The transfer back should be done by the command buffer I believe (either when the texture is registered for use for some other purpose than transfer, of when pipeline barriers get executed for some other reason).

remoe:
The MSL transpiler issue is reported here:

https://github.com/KhronosGroup/MoltenVK/issues/345

remoe:
sure, here is the full spirv/glsl/msl dump:

https://pastebin.com/5GpDQ4uZ

BearishSun:
Thanks. I've taken a quick look, but take this with a grain of salt as I don't really have any experience with MSL.

It might be enough to use `RWStructuredBuffer` instead of `RWBuffer` for the problematic variables. Internally they translate to different things in GLSL:
 - `RWStructureBuffer` -> `buffer` in GLSL (AKA Shared Storage Buffer Object)
 - `RWBuffer` -> `imagebuffer` in GLSL

And when those get translated to MSL:
 - `buffer` -> raw memory in `device` space in MSL
 - `imagebuffer` -> `texture2d` in MSL

Looking at MSL spec, it supports atomic operations for memory in `threadgroup` and `device` spaces (https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf, section 5.13.2), but no such support for `texture2d`. And therefore just switching the buffer types should hopefully be enough to solve the problem.

Note you will also need to edit the C++ code that creates and binds those buffer so it creates a `GpuBuffer` of type `GBT_STRUCTURED`, rather than `GBT_STANDARD`. (Searching for variable names in C++ should reveal locations where they're created/bound).

BearishSun:
Try debugging the point of creation of those buffers, and going all the way down into Vulkan render backend, and see if the sizes passed to Vulkan make sense. If that's not it, then check `VulkanGpuParams::prepareForBind` and see if all the values make sense there as well.

remoe:
Ok, i solved this storage buffer issue. It was because i haven't ported all buffers to "StructuredBuffer". Now i have a Render pipeline error:

Compiler failed to build request
[***MoltenVK ERROR***] VK_ERROR_INITIALIZATION_FAILED: Render pipeline compile failed (error code 1):
fragment shader color output does not have enough component for the pixel format (MTLPixelFormatRGBA16Float).

I'm not sure about this.

BearishSun:
One thing I'm worried about - I'm not sure how (and if) HLSL handles atomic operations on `RWStructuredBuffer`. Any chance you tested the new shaders on Windows/DirectX?

Regarding the new error. Can you track down which pipeline/shader is causing it? I'm guessing it might be something like a shader only writing to RGB, but it should be writing to RGBA, or something like that.

remoe:
About DirectX: The demo runs but i see this on console:

```
D3D11 ERROR: ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 0 of the Compute Shader unit is a Structured Buffer while the shader expects a typed Buffer. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching). [ EXECUTION ERROR #2097379: DEVICE_UNORDEREDACCESSVIEW_BUFFER_TYPE_MISMATCH]
D3D11 ERROR: ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 0 of the Compute Shader unit is using the Format (UNKNOWN). This format does not support 'Add' operations, which are being used on the UAV by the shader. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching). [ EXECUTION ERROR #2097383: DEVICE_UNORDEREDACCESSVIEW_ATOMIC_ADD_UNSUPPORTED]
D3D11 ERROR: ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 1 of the Compute Shader unit is a Structured Buffer while the shader expects a typed Buffer. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching). [ EXECUTION ERROR #2097379: DEVICE_UNORDEREDACCESSVIEW_BUFFER_TYPE_MISMATCH]
D3D11 ERROR: ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 1 of the Compute Shader unit is using the Format (UNKNOWN). This format does not support 'Exchange', which is being used on the UAV by the shader. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching). [ EXECUTION ERROR #2097386: DEVICE_UNORDEREDACCESSVIEW_ATOMIC_EXCHANGE_UNSUPPORTED]
D3D11 ERROR: ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 2 of the Compute Shader unit is a Structured Buffer while the shader expects a typed Buffer. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching). [ EXECUTION ERROR #2097379: DEVICE_UNORDEREDACCESSVIEW_BUFFER_TYPE_MISMATCH]
D3D11 ERROR: ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 2 of the Compute Shader unit is using the Format (UNKNOWN). This format does not support 'Store' which is being used on the UAV by the shader. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching). [ EXECUTION ERROR #2097382: DEVICE_UNORDEREDACCESSVIEW_FORMAT_STORE_UNSUPPORTED]
D3D11 ERROR: ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 3 of the Compute Shader unit is a Structured Buffer while the shader expects a typed Buffer. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching). [ EXECUTION ERROR #2097379: DEVICE_UNORDEREDACCESSVIEW_BUFFER_TYPE_MISMATCH]
D3D11 ERROR: ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 3 of the Compute Shader unit is using the Format (UNKNOWN). This format does not support 'Add' operations, which are being used on the UAV by the shader. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching). [ EXECUTION ERROR #2097383: DEVICE_UNORDEREDACCESSVIEW_ATOMIC_ADD_UNSUPPORTED]
D3D11 ERROR: ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 4 of the Compute Shader unit is a Structured Buffer while the shader expects a typed Buffer. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching). [ EXECUTION ERROR #2097379: DEVICE_UNORDEREDACCESSVIEW_BUFFER_TYPE_MISMATCH]
D3D11 ERROR: ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 4 of the Compute Shader unit is using the Format (UNKNOWN). This format does not support 'Exchange', which is being used on the UAV by the shader. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching). [ EXECUTION ERROR #2097386: DEVICE_UNORDEREDACCESSVIEW_ATOMIC_EXCHANGE_UNSUPPORTED]
D3D11 ERROR: ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 5 of the Compute Shader unit is a Structured Buffer while the shader expects a typed Buffer. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching). [ EXECUTION ERROR #2097379: DEVICE_UNORDEREDACCESSVIEW_BUFFER_TYPE_MISMATCH]
D3D11 ERROR: ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 5 of the Compute Shader unit is using the Format (UNKNOWN). This format does not support 'Store' which is being used on the UAV by the shader. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching). [ EXECUTION ERROR #2097382: DEVICE_UNORDEREDACCESSVIEW_FORMAT_STORE_UNSUPPORTED]
ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 0 of the Compute Shader unit is a Structured Buffer while the shader expects a typed Buffer. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching).
ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 0 of the Compute Shader unit is using the Format (UNKNOWN). This format does not support 'Add' operations, which are being used on the UAV by the shader. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching).
ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 1 of the Compute Shader unit is a Structured Buffer while the shader expects a typed Buffer. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching).
ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 1 of the Compute Shader unit is using the Format (UNKNOWN). This format does not support 'Exchange', which is being used on the UAV by the shader. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching).
ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 2 of the Compute Shader unit is a Structured Buffer while the shader expects a typed Buffer. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching).
ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 2 of the Compute Shader unit is using the Format (UNKNOWN). This format does not support 'Store' which is being used on the UAV by the shader. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching).
ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 3 of the Compute Shader unit is a Structured Buffer while the shader expects a typed Buffer. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching).
ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 3 of the Compute Shader unit is using the Format (UNKNOWN). This format does not support 'Add' operations, which are being used on the UAV by the shader. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching).
ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 4 of the Compute Shader unit is a Structured Buffer while the shader expects a typed Buffer. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching).
ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 4 of the Compute Shader unit is using the Format (UNKNOWN). This format does not support 'Exchange', which is being used on the UAV by the shader. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching).
ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 5 of the Compute Shader unit is a Structured Buffer while the shader expects a typed Buffer. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching).
ID3D11DeviceContext::Dispatch: The Unordered Access View in slot 5 of the Compute Shader unit is using the Format (UNKNOWN). This format does not support 'Store' which is being used on the UAV by the shader. This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching).

		 in auto __cdecl bs::ct::D3D11RenderAPI::dispatchCompute::<lambda_c013ae662fe17d4df567e039195f9650>::operator ()(unsigned int,unsigned int,unsigned int) const [bsf.git\source\plugins\bsfd3d11renderapi\bsd3d11renderapi.cpp:864]```

**Comment by BearishSun** [comment]
Sounds like it's maybe still running the old shader code? You need to compile & run the import tool.

```
cd bsf
mkdir bsfImportToolBuild
cd bsfImportToolBuild
cmake -G "Visual Studio 15 2017 Win64" -DCMAKE_INSTALL_PREFIX=%cd%/../Dependencies/tools/bsfImportTool ..
cmake --build . --config Release --target bsfImportTool
cmake --build . --config Release --target Install
cd ..
```

Then make sure to enable `BUILD_ASSETS` in CMake and re-run it.

remoe:
hm, i have build all from scratch. So, the assets should be correct.

remoe:
FYI: I use now LunarG-SDK v1.1.92. This integrates all important features to build BSF/RobotSample with MoltenVK.

BearishSun:
That crash on resize should be resolved on Linux. You can add similar code to to the mac OS render window. The commit is here: 424828a38a8c47ad3d8b1e07bff43b360bf24bb8.

remoe:
```BSF``` throws the following error on Windows and MacOSX with Vulkan SDK 1.1.92.1:

Description: ERROR: [Validation] Code 0:  [ VUID-VkDescriptorImageInfo-imageLayout-00344 ] Object: 0x184f40b31f0 (Type = 6) | vkCmdDrawIndexed(): Cannot use image 0x1dc (layer=0 mip=0) with specific layout VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL that doesn't match the actual current layout VK_IMAGE_LAYOUT_GENERAL. The Vulkan spec states: imageLayout must match the actual VkImageLayout of each subresource accessible from imageView at the time this descriptor is accessed as defined by the image layout matching rules (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkDescriptorImageInfo-imageLayout-00344)

Need to investigate.

remoe:
The layout issue is solved with a lot of help from @BearishSun ;)

remoe:
It was a pleasure to work together with @BearishSun on this PR. Unfortunately, it took too much time for me. But some code of this PR is usable to integrate and some code is already done better from @BearishSun in current master.

PR #216, opened 2018-10-27
Custom SmallVector implementation

paolopaoletto:


BearishSun:
Great, thanks for doing this. Finally had some time to take a look.

I've looked a bit into the standard library & LLVM implementation of vector and SmallVector respectively, and iterated on your design. The biggest change I think is that they don't call constructors during reallocation. To account for that I've modified the code with `grow` method instead of `realloc`, and moved all the construction outside - also I'm using copy constructor instead of default constructor (also to be more in line to what those two existing implementations do).

I've also updated the code to better handle moves, assignment, support move construction, initializer lists, added some asserts for out of range checks, moved/renamed some stuff for code style purposes, added a unit test and made some fixes.

And finally happy to say the class is now integrated and used by the framework. Nice to have that sorted. If we do a normal vector version I think we can re-use a lot of the code from here.

PR #224, opened 2018-11-14
Allow use of (some) OS libraries

xytovl:
In order to facilitate integration into Linux distributions, it would be preferable to not embed libraries.

This pull request adds a cmake configuration to (not) use bundled libraries:
-DUSE_BUNDLED_LIBRARIES=ON/OFF

FindXXX modules have been split in 3 directories:
OSModules to be used when bundled libraries are deactivated
BundledModules to be used when using bundled libraries
Modules for those that are shared

3 dependencies are still always from the bundle: FBXSDK, PhysX and XShaderCompiler.
Those libraries are not usually part of Linux distributions, so I do not have an immediate solution not to bundle them and still make it easy to compile the project.


BearishSun:
Hi, I haven't had a chance to take a proper look over the PR yet, but just a quick note. bsf should look for system libraries by default if bundled ones are not present - is that behavior not working or somehow inadequate, or does your approach provide some extra functionality? I suppose if required we could also force this behaviour to be enabled through a CMake flag (instead of just searching for the files).

xytovl:
The goal would be to ensure that bundled libraries are not used.

In Linux packaging, we try to avoid as much as possible automatic feature detection. The build should fail if prerequisites are not met.
It makes it easier to provide consistent builds.

An other difference is that by using the modules from the OS itself, it will work with libraries installed in non-standard location, or that add cflags etc.

Finally, it enables bundled mode to copy the libraries, and OS mode to just link to them.

xytovl:
The PR needs some more work, as it broke bundled libraries, I am working on it.

BearishSun:
I understand the reasoning, I'm just saying `bsf` already does that. It first calls `find_path` and `find_library` on the bundled library locations, and if those aren't found, then it calls those same methods using the default search paths - to me this seems like the same thing you are doing in the non-bundled modules? But I might be missing something.

The only difference is that your approach can be toggled with a cmake flag, but we can extend the existing approach with that as well, with no further changes.

PR #225, opened 2018-11-14
Allow deactivation of most bundled libraries

xytovl:
Same goal as #224 but less intrusive.


BearishSun:
This looks good, thanks for making it less intrusive.

PR #229, opened 2018-11-18
Dump HLSL/GLSL/VKSL shaders 

remoe:
This allow to dump HLSL / GLSL / VKSL to the temp folder. This is useful to debug the XCompiler output. Currently it supports VS/FS/CS.


BearishSun:
I'm not confortable integrating this sort of functionality in the importer. I don't like outputting data as a side effect of another operation, even if just for debugging. Plus I usually find it easy enough to set a breakpoint and just read the shader source directly there.

I'd perhaps be fine with it if it was a separate utility class & method for dumping the shader. And ideally this would be something that's only called from the import tool (with a flag that tells it to dump shader data).

remoe:
I understand. I used it to dump all shaders for a XCompiler test.

BearishSun:
XShaderCompiler comes with an optional GUI debugger tool that translates shader code on the fly as you type it, it's quite useful. Requires wxWidgets as a dependency.

PR #230, opened 2018-11-19
Cleanup the build system

Meumeu:
This cleans up a few things in the build system:

* Put the compilation options in target properties to avoid polluting CMAKE_CXX_FLAGS and others
* Install the dependencies in a subdirectory to avoid conflicts with the installed system libraries
* Put the generated headers in the build directory
* Install a cmake file to be able to use find_package
* Don't install .version / .reqversion in the data directory


BearishSun:
Thank you, very nice clean PR.

On Windows I'm having the issue that all the built-in dependency .dlls are still placed in `source/bin` while the framework libraries now end up in `build/bin`. This means that the examples, as well as custom user applications will not run as it won't be able to find the .dlls.

There is a method `copyBsfBinaries` in `HelperMethods.cmake` that's meant to do the .dll copying. Changing that might be enough, but I didn't have time to take a proper look.

BearishSun:
Great, thank you. I'll try it out again in the next few days.

BearishSun:
Thanks for this, everything looks good, merged.

remoe:
RPATH does not work on MacOSX:

```
> otool -l bin/bsfImportTool | grep RPATH -A2

cmd LC_RPATH
cmdsize 40
path $ORIGIN:$ORIGIN/bsf-1.0.0 (offset 12)
```

PR #232, opened 2018-11-26
Feature: Class to container Complex number

paolopaoletto:


BearishSun:
Hey, looks good, nice work. Some of the unit tests fail however. Also some methods have the `Type` template re-declared at function scope, perhaps that should be removed, or renamed to something else if they are meant to represent a different type?

PR #233, opened 2018-11-29
Don't install dependencies in a subdirectory on MacOSX

Meumeu:
This should fix @remoe's issue: this restore the previous behaviour on MacOSX.


remoe:
Hi, i've done the following:

CMake/Properties.cmake

```
if(APPLE)
	set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
else()
	set(CMAKE_INSTALL_RPATH "\$ORIGIN")
endif()
```

CMake/HelperMethods.cmake

```
	if(APPLE)
		set_property(TARGET ${target} PROPERTY INSTALL_RPATH "@loader_path;@loader_path/../lib;@loader_path/bsf-${BS_FRAMEWORK_VERSION_MAJOR}.${BS_FRAMEWORK_VERSION_MINOR}.${BS_FRAMEWORK_VERSION_PATCH}")
	else()
		set_property(TARGET ${target} PROPERTY INSTALL_RPATH "\$ORIGIN:\$ORIGIN/bsf-${BS_FRAMEWORK_VERSION_MAJOR}.${BS_FRAMEWORK_VERSION_MINOR}.${BS_FRAMEWORK_VERSION_PATCH}")
	endif()
```

What you think?

Meumeu:
I guess this should work but I don't have a mac so I can't test it... I'll trust you on this

BearishSun:
Thanks. I've fixed this in a couple of different commits over the last few weeks, as I had some rpath issues on Linux as well. @remoe fix works on macOS. I didn't change install behaviour because a. there are currently no dependency shared libraries used on macOS, and b. even if they were they should be loaded fine with @remoe fix.

Everything appears to work again now, closing this.

Issue #236, opened 2018-12-13
Add Nvidia/AMD driver hint to use high perfomance GPU on laptops with hybrid graphics


DJm00n:
This should help on laptops with hybrid graphics:
```
extern "C" {
_declspec(dllexport) DWORD NvOptimusEnablement = 0x00000001;
}
```
https://docs.nvidia.com/gameworks/content/technologies/desktop/optimus.htm 

```
extern "C" {
_declspec(dllexport) DWORD AmdPowerXpressRequestHighPerformance = 0x00000001;
} 
```
https://community.amd.com/thread/169965

remoe:
Yes, I can confirm this, because I use this on another platform:

```c++
extern "C" {
    __declspec(dllexport) DWORD NvOptimusEnablement = 0x00000001;
    __declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
}
```

PR #237, opened 2018-12-15
 Feature: Add the Discrete GPU force header for use in executables.

guestnone:
This header file when included on final executable is used on Windows-based computers to force the laptops with NVIDIA Optimus or AMD Xpress features to use discrete GPU's
instead of the Intel integrated graphics chip.

Solves issue #236 (at least on Windows, Linux probably works differently).


BearishSun:
Thanks. I need to consider if there's a better way of doing this, I don't like the idea the users need to remember to add a header manually. If I can't think of anything I'll merge.

Issue #252, opened 2019-01-10
Integrate PhysX4?


amerkoleci:
Hi,

Any plans to add PhysX4 support? (https://github.com/NVIDIAGameWorks/PhysX)
It has CMake support and maybe you can build as third party lib into your build system?

Thanks

BearishSun:
Definitely, but it's not a priority right now.

webhaikal:
PhysX has been updated to Physx4. Will create a PR once I've finished testing it. You can follow here: [https://discourse.bsframework.io/t/using-directxtex-as-a-way-to-import-dds-textures/355](https://discourse.bsframework.io/t/using-directxtex-as-a-way-to-import-dds-textures/355)

PR #258, opened 2019-01-16
Feature/physx 4.0

webhaikal:
This PR solves #252. Update PhysX from 3.3 to 4.0, CMake is updated to reflect this but as I'm not familiar with CMake, it might have some errors.

Following on above, you would need to add another value to C++/C->Preprocessor or equivalent the following item:
`PX_PHYSX_CHARACTER_STATIC_LIB`
to avoid linking errors.

I've tested it using the bsfExamples but couldn't make the ball spawn, so I can't test rigidbodies. So you can treat this as WIP. You may also revert commit webhaikal/bsf@318fe2814a5cc298871302c4f590bebde10a5a78


BearishSun:
Thank you. Since this requires me to compile PhysX4 on all platforms, it will take longer than usual to merge, until I get a moment to do so. Let me know when you feel everything is functional on Windows and I'll see about merging.

I believe the stuff in `CMakeFiles` folder was added in error? Could you remove it from the PR?
Why is the documentation download commented out? In any case, it shouldn't be part of the PR.
I notice the event callback is commented out. Is there a replacement mechanism for this?
What's the issue with `eINFLATE_CONVEX`?

webhaikal:
The CMakeFiles & Documentation part was added in by error. Will revert it back. However, you would need to update the FindPhysx CMake file to account for the new naming scheme.

Event Callback was commented out as I am not sure how to implement but will update it around today.

eINFLATE_CONVEX is removed from PhysX 4 as the new algorithm doesn't need to have it inflated.

BearishSun:
Sounds good.

webhaikal:
> Sounds good.

There you go, everything is done, reverted back the documentation comments and updated FindPhysX to reflect new naming scheme. Everything should be functional, just the extra preprocessor flag is needed for bsPhysx `PX_PHYSX_CHARACTER_STATIC_LIB`

PR #260, opened 2019-01-18
Feature: Quadtree implementation.

paolopaoletto:


Source/Foundation/bsfUtility/Utility/BsQuadtree.h
```diff
@@ -0,0 +1,818 @@
...
+
+		/**
+		 * Constructs an quadtree with the specified bounds.
+		 *
+		 * @param[in]	center		Origin of the root node.
+		 * @param[in]	extent		Extent (half-size) of the root node in all directions;
+		 * @param[in]	context		Optional user context that will be passed along to getBounds() and setElementId()
+		 *							methods on the provided Options class.
+		 */
+		Quadtree(const Vector3& center, float extent, void* context = nullptr)
```
BearishSun: Ideally the center is a `Vector2` as well.

BearishSun:
Looks good, thanks for taking the time to make a 2D version. I'd appreciate if we can change the bounds to `Rect2` as I mentioned, but it's not a requirement. To use it you'd probably need to extend `Rect2` it a bit, as well as create a SIMD version similar to `AABox`.

PR #266, opened 2019-01-24
Feature: Rect2 extended and SIMD implmentation.

paolopaoletto:


BearishSun:
Thanks. `Rect2` additions weren't calculating the center and half-size properly, I modified that bit locally. Merged.

PR #269, opened 2019-02-09
Feature: Quadtree implementation and Quadtree utility test suite.

paolopaoletto:


BearishSun:
The issue was the unit test. `Rect2` constructor was accepting 'position + size' in the size field, making the rects very large, and outside of the 800 unit extent of the root node. Fixing that solved the crash - ultimately we might want a better error if a node is outside the range, but I don't have time for that at the moment.

Anyway, it's merged. Thanks!

PR #271, opened 2019-02-15
Feature: Vector3I and Vector4I for bsfSharp scripting.

paolopaoletto:


BearishSun:
Thanks, merged. I've removed the stuff I don't think is useful for integer vectors, like normalization, orthogonal base generation, cross product and similar. Also made the classes partial so they extend the auto-generated integer vector types (among other things, this ensures we don't accidentally break the data layout between C++ and C#).

PR #272, opened 2019-02-16
Rename BsRect3.cs to Rect3.cs

paolopaoletto:


BearishSun:
Good find, thanks, merged.

PR #273, opened 2019-02-19
Feature: Mix and Max method with variadic template alternative

paolopaoletto:


Source/Foundation/bsfUtility/Math/BsMath.h
```diff
@@ -521,7 +521,7 @@ namespace bs
 		}
 
 		/** Returns the minimum value of all the values provided. */
-		template<typename T, typename... Args>
+		template <typename T, typename... Args>
```
jonesmz: Is there a way to reduce the depth of the recursive function calls?

Something like

`return min(min(a, b), min(std::forward<Args>(args)...));`

?

Then instead of N^2 depth, you get linear depth, I think (Not spending a lot of time on this comment to verify my math, sorry).

You'll want to provide an overload of min that takes only a single argument and just returns it, though. That way if you have an odd number of arguments, it still works.

There's probably another way to handle it to get even smaller amounts of recursion, but I'm going to leave that for other people to noodle over.

jonesmz: Note: This is mostly a compile-time concern. The actual runtime of the function probably won't be effected.

jonesmz: Another thought is to use std::common_type_t, so that the input types don't all have to be identical. 

```
template<typename T>
const T & min(const T & in)
{
    return in;
}

template<typename A_T, typename B_T>
std::common_type_t<A_T, B_T> min(const A_T & a, const B_T & b)
{
    if(a < b)
    {
        return a;
    }
    else
    {
        return b;
    }
}

template<typename A_T, typename B_T, typename ... ARGS_T>
std::common_type_t<A_T, B_T, ARGS_T...> min(const A_T & a, const B_T & b, const ARGS_T & ... args)
{
    return min(min(a, b), min(args...));
}
```

BearishSun:
Thank you, useful functions to have. @jonesmz made some good suggestions so I adjusted the code to match them. Merged.

jonesmz:
By the way, if BSF ever adopts C++17, instead of C++14, you could do it this way.

```
template<class FIRST_T, class ... ARGS_T>
std::common_type_t<A_T, ARGS_T...> min(FIRST_T a, ARGS_T const& ... args)
{
    return (a1, ..., (a1 = args < a ? args : a));
}
```

Modified from this paper http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1219r1.html#MotivationAndScope

PR #277, opened 2019-02-26
Feature: SmoothStep method and ApproxEquals methods with different ty…

paolopaoletto:
…pes for bsfSharp math.


BearishSun:
Looks good, thank you. Merged.

PR #278, opened 2019-03-07
Feature: Interpolation methods and various fast utility methods.

paolopaoletto:


BearishSun:
Can you elaborate a bit on the `quintic` and `cubic` functions? What their parameters are and how are they meant to be used?

BearishSun:
Thanks for the clarification, looks good. Merged.

PR #286, opened 2019-03-17
Feature: Bitwise for bsfSharp

paolopaoletto:


BearishSun:
Hey. I'm getting multiple compile errors in this one.

BearishSun:
Thanks. I've merged this partially. I left out the uint <-> snorm/unorm and integer lerp/invLerp methods. 

I'd need tests in order to confirm those still work as intended, as there were some uint -> int changes that might have changed the behavior. Additionally those are used in a very very specific bits for C++ code and the use in C# code might be very limited. In general C# isn't a language I see much bit manipulation. But regardless if you take the time those are tested and work the same in C++ I'll merge them as well.

PR #293, opened 2019-03-25
Feature: DenseMap utility implementation

paolopaoletto:


BearishSun:
Nice work, thanks for adding the documentation. 

I am getting a compile error on this line:
`			mCount = size() > 32 ? 1 << (std::log(size()) + 1) : 64;`

> Cannot apply binary << to 'int' and 'double'.

BearishSun:
Also I noticed the code is mixing `bs_allocN` and `bs_delete`, `allocN` will allocate the memory without calling the constructor, while `bs_delete` will call the destructor and then free the memory.

`bs_alloc`/`bs_allocN` should be freed through `bs_free`, while `bs_construct` should be freed with `bs_delete`.

Not sure whether the intent was to call the constructor/destruction or not...

BearishSun:
Thanks for making the changes. Awesome work on this one. Merged.

PR #297, opened 2019-03-31
Feature: Random class for bsfSharp math

paolopaoletto:


BearishSun:
There is already a `Random` class in bsfSharp. I'd prefer to keep that one instead of duplicating code too much - especially when we add more scripting languages that will prove a real hassle to translate.

PR #299, opened 2019-04-08
Feature: Smart pointer that does not support usage across multiple th…

paolopaoletto:
…reads


BearishSun:
Thanks for adding this one! I've been meaning to get around to it but never have the time. It'll very much come in handy. I did change the implementation a bit and hopefully simplified a few parts.

Later on I'd love to have `make_shared` for it, ensuring both the pointer and the counter can be allocated in a single allocation, instead of needing two separate ones.

Anyway, merged.

PR #312, opened 2019-04-29
Misc build fixes for MSVC Ninja build

ReDucTor:
By default 'Open Folder' in MSVC uses the Ninja generator, this fixes some issues when it attempts to run and simplifies some code

The other notable change in here is changing RelWithDebInfo to use -DNDEBUG not -DDEBUG to match the default CMake environment, this also fixes issues where the C# Assembly loading is expecting to use the Debug path when using RelWithDebInfo.

### DIFF
--- a/Source/CMake/HelperMethods.cmake
+++ b/Source/CMake/HelperMethods.cmake
@@ -109,14 +109,8 @@ MACRO(add_imported_library LIB_NAME RELEASE_NAME DEBUG_NAME IS_SHARED)
 		add_library(${LIB_NAME} STATIC IMPORTED)
 	endif()
 
-	if(CMAKE_CONFIGURATION_TYPES) # Multiconfig generator?
-		set_target_properties(${LIB_NAME} PROPERTIES IMPORTED_LOCATION_DEBUG "${DEBUG_NAME}")
-		set_target_properties(${LIB_NAME} PROPERTIES IMPORTED_LOCATION_RELWITHDEBINFO "${RELEASE_NAME}")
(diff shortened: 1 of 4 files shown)


BearishSun:
Thanks for looking into this.

I have some concerns with `RelWithDebInfo`, as it is now identical to `Release`. We generate debug symbols for both configurations and that flag was the bit that made them different. It's not a configuration that I use much personally, but regardless it doesn't feel right to have them be identical. If we can fix the C# project in some other way I'd be happier.

ReDucTor:
Standard CMake `Release` normally doesn't include generation of debug info. Another solution if you want to avoid the identical builds is to instead make it so there is only 'Release' and 'Debug' configurations, eliminating the other standard ones for CMake.

BearishSun:
Eliminating or changing the default `CMake` configurations caused problems for people in the past, generally when it comes to importing the project into other CMake projects which then don't have the same set of configurations, causing problems.

Symbols with release binaries I think are always useful so I've enabled them. And I think it's also useful to have an optimized build with all the internal debugging mechanisms enabled (which is currently `RelWithDebInfo`).

If nothing else works, perhaps we can just add an extra flag that forces the runtime to look for assemblies in a different spot for that particular configuration (assuming that was the only issue). As a quick solution, I can merge the rest of the PR if we remove that bit and handle that later, if you agree.

ReDucTor:
Sure, cherry picking parts of it, seems good, the primary issue that was trying to fix was loading debug assemblies with RelWithDebInfo, however looking deeper it appears there is a few places where this can happen, along with loading from build directory over the install directory.

I'll move my environment over to use 'Release' by default instead of 'RelWithDebInfo' if thats the more expected/stable usage for this project.

BearishSun:
Sounds good, merged with the discussed modification, thanks.

PR #313, opened 2019-04-29
 Build: Versioning and resource file improvements

guestnone:
This PR adds the autogenerated classes for getting the version of the framework and engine/editor from the C# side and exposes the engine/editor version macros on C++ side.
It (should) also fixes the problem introduced after the script revamp when Mono script runtime resource file would be linked with the Banshee executables, while it should only linked with the bs::f one.


BearishSun:
Looks good thanks, especially the fix for Banshee executable getting the wrong info. Merged.

PR #314, opened 2019-04-30
Don't add all files to 'bin' directory for Banshee3D

ReDucTor:
NOTE: This does not(should not) change the bsf install


ReDucTor:
Did a quick check of the output from appveyor doesn't look like anything changed for bsf, hopefully just for b3d

BearishSun:
Looks like it should be fine, thank you. Merged.

PR #317, opened 2019-05-01
Provide unmapped triangle index for mesh raycast queries

mworchel:
This PR adds the member `PhysicsQueryHit::unmappedTriangleIndex` which allows using the result of a raycast query to index faces in the original mesh and not only in the physics mesh associated with a mesh collider. It is useful for cases where arbitrary vertex semantics are to be evaluated (e. g. vertex color) which are not accessible through the physics mesh (as it only contains spatial vertex information).


BearishSun:
Thank you, very useful to have. Merged.

PR #319, opened 2019-05-05
Use RTTIFieldInfo::DEFAULT with explicit namespace in macros

mworchel:
This PR adds the explicit namespace `bs` to the occurences of `RTTIFieldInfo::DEFAULT` in macros. A code (only the relevant bits) like
```cpp
namespace MyNamespace
{
  class A : bs::Component
  {
    ...
    int mMember;
  }

  class RTTI_A : public bs::RTTIType<A, bs::Component, RTTI_A>
  {
    BS_BEGIN_RTTI_MEMBERS
    BS_RTTI_MEMBER_PLAIN(mMember, 0)
    BS_END_RTTI_MEMBERS
    
    ...
  };

} 
```
would not compile before, as macros are placed in global namespace and `BS_RTTI_MEMBER_PLAIN(mMember, 0)` would resolve to a snippet containing `RTTIFieldInfo::DEFAULT`. For this to compile, it would require `using namespace bs` but that is obviously not desired for header files.

### DIFF
--- a/Source/Foundation/bsfUtility/Reflection/BsRTTIType.h
+++ b/Source/Foundation/bsfUtility/Reflection/BsRTTIType.h
@@ -55,10 +55,10 @@ namespace bs
 	 * The type of the member must be a valid plain type. Each field must specify a unique ID for @p id.
 	 * An optional @p RTTIFieldInfo structure can be provided to provide further information about the field.
 	 */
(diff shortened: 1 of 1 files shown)


BearishSun:
Yep, makes sense. Thanks, merged.

ataulien:
There are also some occurrences of `bs::UINT32` which are without namespace when using the Array-variants. Also `BS_EXCEPT` doesn't work outside the bs namespace for that reason.

PR #320, opened 2019-05-05
Use struct as deleter for smart pointers by default

mworchel:
This PR adds a callable `Deleter` struct that serves as a proxy for the `bs_delete` function and is passed by default to the `bs::UPtr` and `bs::SPtr` types. It replaces the use of function pointers to `bs_delete` for these types. Since the new deleter is default-constructible it now allows default-constructible smart pointers like
```cpp
UPtr<A> foo;
SPtr<A> foo;
```
A deleter of a base type is also constructible from a deleter of a derived type, which allows smart pointer polymorphism 
```cpp
class A {};
class B : public A {};

UPtr<A> foo = bs_unique_ptr_new<B>();
SPtr<A> bar= bs_unique_ptr_new<B>();
```
With these changes the bs::f smart pointers behave more like the standard smart pointers while still following the custom allocator architecture.

### DIFF
--- a/Source/Foundation/bsfUtility/Allocators/BsMemoryAllocator.h
+++ b/Source/Foundation/bsfUtility/Allocators/BsMemoryAllocator.h
@@ -269,6 +269,22 @@ namespace bs
 		MemoryAllocator<Alloc>::free(ptr);
 	}
 
+	/** Callable struct that acts as a proxy for bs_delete */
+	template<class T, class Alloc = GenAlloc>
+	struct Deleter
+	{
+		constexpr Deleter() noexcept = default;
+
+		/** Constructor enabling deleter conversion and therefore polymorphism with smart points (if they use the same allocator). */
+		template <class T2, std::enable_if_t<std::is_convertible_v<T2*, T*>, int> = 0>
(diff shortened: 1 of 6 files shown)


Source/Foundation/bsfUtility/Allocators/BsMemoryAllocator.h
```diff
@@ -269,6 +269,22 @@ namespace bs
...
 	}
 
+	/** Callable struct that acts as a proxy for bs_delete */
+	template<class T, class Alloc = GenAlloc>
+	struct Deleter
+	{
+		constexpr Deleter() noexcept = default;
+
+		/** Constructor enabling deleter conversion and therefore polymorphism with smart points (if they use the same allocator). */
+		template <class T2, std::enable_if_t<std::is_convertible_v<T2*, T*>, int> = 0>
```
mworchel: Oops, that is c++17. Currently fixing it.

BearishSun:
Great work, makes unique pointers much more usable. I modified `DynLib` code so it doesn't require custom deleters as they made the code ugly.

Merged.

PR #325, opened 2019-05-09
Guielement

jayrulez:
For matching this: https://github.com/GameFoundry/bsf/blob/master/Source/Scripting/bsfScript/Wrappers/GUI/BsScriptGUIElement.cpp#L114


BearishSun:
Good catch, thanks.

PR #327, opened 2019-05-12
Copy files from correct directory

jayrulez:
Make it so files are copied from {srcSir}/bin insread of {srcSir}/bin/bin
Tested and working on windows.

I did the same change for mac but I'm not able to test on mac until this weekend.


BearishSun:
Good catch, thanks.

PR #328, opened 2019-05-14
Use 'correct' types for thunks and set mOnResetThunk to nullptr when …

jayrulez:
…clearing existing


BearishSun:
Looks good, thanks.

PR #329, opened 2019-05-14
Feature: RTTI-PlainType for std::list

ataulien:
Adds a Plain-Type for `std::list` aka `bs::List`, so you can simply use `BS_RTTI_MEMBER_PLAIN` on member variables of those types.

### DIFF
--- a/Source/Foundation/bsfUtility/Prerequisites/BsRTTIPrerequisites.h
+++ b/Source/Foundation/bsfUtility/Prerequisites/BsRTTIPrerequisites.h
@@ -263,6 +263,76 @@ namespace bs
 	};
 
 	/**
+	 * RTTIPlainType for std::list.
+	 *
+	 * @see		RTTIPlainType
+	 */
+	template<class T> struct RTTIPlainType<std::list<T, StdAlloc<T>>>
+	{
+		enum { id = TID_List }; enum { hasDynamicSize = 1 };
+
+		/** @copydoc RTTIPlainType::toMemory */
+		static void toMemory(const std::list<T, StdAlloc<T>>& data, char* memory)
+		{
+			UINT32 size = sizeof(UINT32);
+			char* memoryStart = memory;
+			memory += sizeof(UINT32);
+
+			UINT32 numElements = (UINT32)data.size();
+			memcpy(memory, &numElements, sizeof(UINT32));
+			memory += sizeof(UINT32);
+			size += sizeof(UINT32);
+
+			for(const auto& item : data)
+			{
+				UINT32 elementSize = rttiGetElemSize(item);
+				RTTIPlainType<T>::toMemory(item, memory);
+
+				memory += elementSize;
+				size += elementSize;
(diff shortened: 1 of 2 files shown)


Source/Foundation/bsfUtility/Prerequisites/BsRTTIPrerequisites.h
```diff
@@ -262,6 +262,76 @@ namespace bs
...
+			for(const auto& item : data)
+			{
+				UINT32 elementSize = rttiGetElemSize(item);
+				RTTIPlainType<T>::toMemory(item, memory);
+
+				memory += elementSize;
+				size += elementSize;
+			}
+
+			memcpy(memoryStart, &size, sizeof(UINT32));
```
ReDucTor: While its following the pattern of the other containers, do these size prefixes add any value?

Source/Foundation/bsfUtility/Prerequisites/BsRTTIPrerequisites.h
```diff
@@ -262,6 +262,76 @@ namespace bs
...
+				UINT32 elementSize = RTTIPlainType<T>::fromMemory(element, memory);
+				data.push_back(element);
+
+				memory += elementSize;
+			}
+
+			return size;
+		}
+
+		/** @copydoc RTTIPlainType::toMemory */
```
ReDucTor: This should probably copy doc from getDynamicSize

Source/Foundation/bsfUtility/Prerequisites/BsRTTIPrerequisites.h
```diff
@@ -262,6 +262,76 @@ namespace bs
...
+				RTTIPlainType<T>::toMemory(item, memory);
+
+				memory += elementSize;
+				size += elementSize;
+			}
+
+			memcpy(memoryStart, &size, sizeof(UINT32));
+		}
+
+		/** @copydoc RTTIPlainType::toMemory */
```
ReDucTor: This should copy doc from fromMemory

Source/Foundation/bsfUtility/Prerequisites/BsRTTIPrerequisites.h
ataulien: Fixed that for the other containers, too.

Source/Foundation/bsfUtility/Prerequisites/BsRTTIPrerequisites.h
ataulien: I was wondering the same as it's not really used in the `fromMemory`-functions. Maybe it's just that *all* saved types have it, so it's better to stay consistent, but @BearishSun should be able to say more about that.

BearishSun: Sizes must be written for all dynamically sized plain types, it won't work otherwise.

Issue #331, opened 2019-05-15
Endless audio stream support


frabert:
This is a followup from https://discourse.bsframework.io/t/endless-audio-stream-support/434

I'm willing to work on adding this feature to the codebase. I am thinking that the best way would _not_ be to modify `AudioClip` to support "infinite" lenghts, but to introduce a new class (`AudioStream` maybe?) that only supports PCM audio from a stream.What do you think?

BearishSun:
It would be really nice to have support for this.

I would only consider doing it in a separate class if the solution requires a radically different approach than what `AudioClip` already does. I took a quick look at it when I answered that post on discourse, and it seemed like only minor modifications regarding checks for fixed length might be required. `AudioClip` already supports different load modes, so adding one more wouldn't be a stretch, even if the modifications aren't as simple.

Compare that to adding a separate class, things get a lot more complex as they need to be usable in the same places (meaning same base, which breaks the API, or deriving from `AudioClip` itself, which will probably be a lot of refactoring). Also the plugin implementations don't like implementing class hierarchies, requiring intermediate classes which make things even more complex, esp. if a simple `if` could have done the job.

frabert:
Interesting, because the idea of creating a separate class was _precisely_ to not introduce breaking changes for previous users 😄 I will take a look at it and report back when I know something more about it

PR #333, opened 2019-05-19
[WIP] Endless audio clips

frabert:
I've started working on a fix for #331 

This is a work in progress, I'm creating the PR to show progress and to receive/provide feedback on the solutions implemented.

Currently, only the OpenAL backed is implemented, and it has some issues. Mainly, the playback does not seems "fluid", I'm guessing the worker thread is not queueing the audio buffers fast enough, and OpenAL automatically stops the source, so that it has to be restarted manually every time a new buffer is enqueued.

### DIFF
--- a/Source/Foundation/bsfCore/Audio/BsAudio.h
+++ b/Source/Foundation/bsfCore/Audio/BsAudio.h
@@ -92,7 +92,7 @@ namespace bs
 			const AUDIO_CLIP_DESC& desc) = 0;
 
 		/** 
-		 * Creates a new audio clip.
+		 * Creates a new audio clip, for use with endless data streams.
 		 *
 		 * @param[in]	samples		Stream containing audio samples in format specified in @p desc.
 		 * @param[in]	desc		Descriptor describing the type of the audio stream (format, sample rate, etc.).
--- a/Source/Foundation/bsfCore/Audio/BsAudio.h
+++ b/Source/Foundation/bsfCore/Audio/BsAudio.h
@@ -91,6 +91,15 @@ namespace bs
 		virtual SPtr<AudioClip> createClip(const SPtr<DataStream>& samples, UINT32 streamSize, UINT32 numSamples,
 			const AUDIO_CLIP_DESC& desc) = 0;
 
+		/** 
+		 * Creates a new audio clip.
+		 *
+		 * @param[in]	samples		Stream containing audio samples in format specified in @p desc.
+		 * @param[in]	desc		Descriptor describing the type of the audio stream (format, sample rate, etc.).
+		 * @return					Newly created AudioClip. Must be manually initialized.
+		 */
+		virtual SPtr<AudioClip> createClip(const SPtr<DataStream>& samples, const AUDIO_CLIP_DESC& desc) = 0;
+
 		/** Creates a new AudioListener. */
 		virtual SPtr<AudioListener> createListener() = 0;
 
--- a/Source/Plugins/bsfOpenAudio/BsOAAudioSource.cpp
+++ b/Source/Plugins/bsfOpenAudio/BsOAAudioSource.cpp
@@ -582,8 +582,10 @@ namespace bs
 
 			if (fillBuffer(mStreamBuffers[i], info, totalNumSamples))
 			{
-				for (auto& source : mSourceIDs)
+				for (auto& source : mSourceIDs) {
 					alSourceQueueBuffers(source, 1, &mStreamBuffers[i]);
+					alSourcePlay(source);
+				}
 
 				mBusyBuffers[i] |= 1 << i;
 			}
@@ -634,7 +636,7 @@ namespace bs
 		bs_stack_free(samples);
 
 		// If the number of read samples is less the the amount we requested, we've reached the end of the stream.
-		return samplesRead == numSamples;
+		return samplesRead == sampleBufferSize;
 	}
 
 	void OAAudioSource::applyClip()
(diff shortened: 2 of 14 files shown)


BearishSun:
Hey. At first glance looks pretty good! Not sure about the buffering issue, it's been a while since I've been in that codebase but if you have trouble figuring it out let me know. I'll leave some comments on the other things I noticed.

Source/Foundation/bsfCore/Audio/BsAudio.h
```diff
@@ -91,6 +91,15 @@ namespace bs
...
 			const AUDIO_CLIP_DESC& desc) = 0;
 
+		/** 
+		 * Creates a new audio clip.
+		 *
+		 * @param[in]	samples		Stream containing audio samples in format specified in @p desc.
+		 * @param[in]	desc		Descriptor describing the type of the audio stream (format, sample rate, etc.).
+		 * @return					Newly created AudioClip. Must be manually initialized.
+		 */
+		virtual SPtr<AudioClip> createClip(const SPtr<DataStream>& samples, const AUDIO_CLIP_DESC& desc) = 0;
```
BearishSun: Documentation might note here it's for creating an endless audio clip. Alternatively we could maybe use the constructor above but assume endless if streamSize and numSamples is 0. Not sure if that's a better option though, probably doesn't matter much.

Source/Plugins/bsfOpenAudio/BsOAAudioSource.cpp
```diff
@@ -492,7 +497,16 @@ namespace bs
 		info.sampleRate = mAudioClip->getFrequency();
 		info.numSamples = 0;
 
-		UINT32 totalNumSamples = mAudioClip->getNumSamples();
+		UINT32 totalNumSamples;
+
+		if(mAudioClip->isEndless())
```
BearishSun: Code style: Omit `{}` for single line blocks

Source/Plugins/bsfOpenAudio/BsOAAudioSource.cpp
```diff
@@ -568,8 +582,10 @@ namespace bs
 
 			if (fillBuffer(mStreamBuffers[i], info, totalNumSamples))
 			{
-				for (auto& source : mSourceIDs)
+				for (auto& source : mSourceIDs) {
```
BearishSun: Code style: `{` should be on a new line

Source/Plugins/bsfOpenAudio/BsOAAudioSource.cpp
```diff
@@ -568,8 +582,10 @@ namespace bs
...
 					alSourceQueueBuffers(source, 1, &mStreamBuffers[i]);
+					alSourcePlay(source);
```
BearishSun: This is the issue you mentioned, that we're forcing OpenAL to start playing again?

Source/Plugins/bsfOpenAudio/BsOAAudioSource.cpp
```diff
@@ -597,18 +623,20 @@ namespace bs
 		UINT32 sampleBufferSize = numSamples * (info.bitDepth / 8);
 
 		UINT8* samples = (UINT8*)bs_stack_alloc(sampleBufferSize);
+		std::fill(samples, samples + sampleBufferSize, 0);
```
BearishSun: Ideally we can avoid doing a fill over the entire buffer. It wastes performance as in 99% of the cases it will just get overwritten.

Source/Plugins/bsfOpenAudio/BsOAAudioSource.cpp
frabert: Yes, ideally the source should not stop playing. In this case, I have to force the source to keep playing, but I am not yet sure why. As I mentioned before, I suspect it is due to the source running out of buffer contents too early because the audio thread doesn't keep up, but I don't know how to test this theory as of now...

frabert:
I have addressed the stilistic issues

frabert:
I've hit a roadblock in that I could not find an easy way to implement this feature using the FMOD backend, but this is probably mainly due to my inexperience with its API. @BearishSun Do you have any pointers on this?

BearishSun:
Current FMOD streaming setup is done in `FMODAudioClip::createStreamingSound`. It sets up an `FMOD_CREATESOUNDEXINFO` struct with an `pcmReadCallback` which gets called by FMOD when it needs new samples - there might be some modifications needed there, not sure. It also has a `length` parameter which probably needs to be set to 0. Potentially some other params there need changing as well, again not sure.

frabert:
Thanks for the tips! Seems like the FMOD backend mostly works now, but the `length` field cannot be 0 (it will make `createSound` fail otherwise), so for now I've just set it to the biggest possible value. This leads me to the question of how to signal that a stream is actually ended: I still don't know, since returning an `EOF` error when running out of data doesn't seem to make playback stop, just begins a "looping" phase

BearishSun:
Looking at this thread: https://qa.fmod.com/t/notifying-user-created-stream-of-end/11417/2

Seems like there is no built-in way to stop an infinite stream. What you could probably do, is to set an atomic in the stream read callback when it ends. The main thread (AudioManager?) can then check the atomic if the stream ended, and call on FMOD to stop playback. Since there is some delay between the two, you can probably fill the stream with zeroes until the main thread actually stops it.

BearishSun:
(Although that thread is also quite old. It might be worth checking FMOD API reference for something newer)

BearishSun:
FMOD also supports playing DSP's, which can generate PCM samples and don't have the limitation of having to specify the length. `api/lowlevel/examples/generate_tone.cpp` example provided in FMOD SDK is a good example.

I'm not sure if that's much better than the sound approach though.

PR #336, opened 2019-05-24
Added externalWindowHandle support for Linux

hhyyrylainen:
This is a pretty basic fix for my problem of not being able to render to an existing window on Linux. 

This also most likely breaks input on external
windows as I left a bunch of the X related setup out because I don't
want bsf to do anything extra to my window. I use SDL2 for input, so I don't need that. A
proper solution to this is probably having one more option for
specifying if bsf should run the window event loop or not, Windows
version probably also needs this option to not steal events from SDL, but I haven't tested my code
on Windows yet.


BearishSun:
Thanks for adding this. Merged in b5bbf3efbe0c79e35362cc2ca5b1368afab9f42c.

PR #341, opened 2019-06-01
fix sse, UtilityTest

pgruenbacher:
I don't actually really expect this to be merged but is meant for comparison to highlight some issues and fixes.

1. The BUILD_TESTS works if I add -msse4.1 to my cmakelist. 
2. UtitliyTest has an infinite recurisve ERR due to modules not being setup, so i added the minimal modules. Likewise MemStack::beginThread was necessary. 
3. StaticVector doesn't seem to work, so I took them out and just did std::vector::reserve instead. I feel like Static allocations are a bit of an overoptimization compared to just doing basic reserve operations in general, it just adds fragility.


pgruenbacher:
looking at the unit tests and test suite...
I think the whole thing should just be dropped and reworked with gtest and should interface with `make test`. The bsf ASSERT macro(s) aren't useful nor is the console output really useful. https://github.com/google/googletest

BearishSun:
I agree, the test system is an afterthought that never had any proper development time invested in it. It's just something that I can run before doing a release and make sure some less used features still work. Ultimately we should change it and hook it up with CI, but it's not high priority right now.

Thanks for providing these fixes. I'll try to integrate them soon.

pgruenbacher:
I'll keep working on this fork branch to migrate everything to gtest and I may also try to integrate https://github.com/google/benchmark for benchmarking. I didn't see any benchmarks (for the non-rendering side of things anyways), which would be good for justifying the custom memory allocations and for testing the acceleration structures.

pgruenbacher:
what's your thoughts on c++17? It'd be def useful for the std::filesystem replacing the custom filesystem platform code and for the standard library parallel algorithms. I don't know how other users feel about upgrading but the code compiles fine for me on gcc-9 so there probably isn't too much issue with upgrading.

BearishSun:
I want to upgrade. It's been on my agenda for a while, but there has always been something higher priority. It shouldn't be much work, mostly just confirming everything still runs/compiles on other platforms.

pgruenbacher:
ok so I've finished updating all the tests to gtest. It's a lot more better when running ./bin/...../unit_tests I think now, and there's a lot more test macros available. Plus all tests just need to have `_test.cpp` suffix and they'll be automatically compiled, no need to declare the filenames.

pgruenbacher:
@BearishSun I'm having an issue trying to get bsModule working with dynamic libraries. I'm getting bsfEnTT as a Plugin shared object. I'm trying to do ECSManager::instance()->getRegistry() but ECSManager::startUp() is leading to multiple singletons. I'm not sure how to get around this....

BearishSun:
Where is `ECSManager::startUp()` getting called? And how is the issue manifesting itself?

There are existing uses of `Module` in various plugins.

pgruenbacher:
it was getting called by the ECSRenderer during the initializeCore() where other modules like RendererUtility were being called.

BearishSun:
How is the issue manifesting itself. i.e. what error are you getting?

pgruenbacher:
I figured out a work-around. It makes sense why it wasn't working for me. sorry for distraction. Once I've gotten a working demo we can worry about how we want to actually organize everything on a module level I guess.

pgruenbacher:
I was trying to introduce screen capture functionality into the unit-tests to better the unit-testing
by following the offscreen-rendering manual but it looks like the api is out of date.

HSceneObject cameraSO = SceneObject::create("Camera");
HCamera camera = cameraSO->addComponent<CCamera>(renderTexture);

this ^ doesn't seem to compile. The fix isn't obvious ot me since the ct::Camera can be constructed with a renderTarget but I'm not sure the best way to get teh CCAmera to construct the core camera with the render target via forwarding...

BearishSun:
Indeed the API changed, I'll update that. This is the correct way:
```
	HCamera sceneCamera = sceneCameraSO->addComponent<CCamera>();
	sceneCamera->getViewport()->setTarget(renderTexture);
```

pgruenbacher:
ahh that is much more sensible than constructor forwarding.

pgruenbacher:
@BearishSun you got code for saving textures to png format? 
this doesn't work out the gate 
```
    	gResources().save(renderTexture->getColorTexture(0), "test.png", overrwrite);
```
but I couldn't find any api for writing to PixelData and then to image or anything like that either...

BearishSun:
` Debug::writeAsBMP` is the closest that you'll find, there are no image saving functions other than that.

pgruenbacher:
hm yea that's gonna be more trouble than it's worth. ok well i guess unit-tests can just be manually tested for the actual visualization.

PR #342, opened 2019-06-04
Add toString for vector classes

JohnathanFL:
Add a toString method for all VectorXT classes, following the format '<X, Y, ...>'.

If there's any interest/this is accepted at all I can add some for the other math classes.

### DIFF
--- a/Source/Foundation/bsfUtility/Math/BsVector2.h
+++ b/Source/Foundation/bsfUtility/Math/BsVector2.h
@@ -342,6 +342,12 @@ namespace bs
 			return Math::isNaN(x) || Math::isNaN(y);
 		}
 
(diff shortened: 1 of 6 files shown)


BearishSun:
The current approach is to use the global `toString` specialization for all to-string conversion needs, which currently supports various vector types (although I think some such as `Vector3I` and `Vector4I` aren't supported).

JohnathanFL:
Woops, sorry. I got an error message when I originally tried to format a vector3 with StringFormat saying that std::to_string didn't support it, and saw that StringFormat's toString method just redirected to std::to_string, so I didn't think to look for another method.

BearishSun:
I don't think you can print vectors through `StringFormat` currently, although that would be a good addition.

PR #350, opened 2019-06-16
Exposed SceneObject hasFalg method to C#

jayrulez:


BearishSun:
Looks good, thanks. Merged, and I've updated b3d with latest bsf so it's reflected there.

PR #357, opened 2019-07-05
Refactor: Log Improvements

guestnone:
PR Outline
----------
This PR Improves and fixes some things to the logging system following its refactorization. It also adds some features to the system itself.

String to time::t conversion
----------------------------
A new function called bs_convert_timet_to_string was added to conveniently convert C++ Standard library
time::t function to the String. The user can choose if the format should be compatible with the ISO 8601
standard or use the formatting used in the HTML log. All functions in bs::Time class that outputted time to string
were moved to use this function.

Text Log
--------
The bs::framework user can now write the log as a standard text file instead of the HTML based one. This should lessen the use of the web browser to read the log, especially for terminal users. The saveLog function can also write the log based on its type.

~~~~~~~~~~~~~~~~~
This is bs::framework 1.1.0
Started on: Friday, July 05, 2019 22:19:26

System information:
================================================================================
OS version: Windows 10 64-bit
CPU information:
CPU vendor: GenuineIntel
CPU name: Intel(R) Core(TM) i5-4200H CPU @ 2.80GHz
CPU clock speed: 2794Mhz
CPU core count: 4

GPU List:
================================================================================
GPU #0: Intel(R) HD Graphics 4600
GPU #1: NVIDIA GeForce GTX 950M
GPU #2: Microsoft Basic Render Driver

Log entries:
================================================================================
2019-07-05T22:19:28Z [INFO] <Uncategorized> | Test message from Info channel
2019-07-05T22:19:28Z [WARNING] <Uncategorized> | Test message from Warning channel
2019-07-05T22:19:48Z [ERROR] <Uncategorized> | Test message from Error channel at 22:19:48
2019-07-05T22:19:48Z [INFO] <Uncategorized> | Friday, July 05, 2019 20:19:48
2019-07-05T22:19:48Z [WARNING] <Uncategorized> | Test Message send using BS_LOG on the Warning channel
                                                  in int __cdecl bs_main(int,char *[]) [D:\XXX\Logging\Main.cpp:59]

2019-07-05T22:19:48Z [FATAL] <Platform> | Something Exploded
~~~~~~~~~~~~~~~~~

Log Name Categorization
-----------------------
A special module called LogCategoryName was added, so that the log category name could be converted to the string. This allows the saved log to display the module from which the message was written.

Other
-----
- Fixed the problem with the time in HTML Log being outputted incorrectly.
- Added the toString() function for LogVerbosity
- Documentation was fixed to cover the logging system revamp.

Notes
-----
This PR wasn't tested on Linux and when compiled as a Banshee Engine.


BearishSun:
Hi, great PR, thank you.

Regarding `bs_convert_timet_to_string` I'd prefer it's called `toString`, so it is consistent with other string converting methods in the framework. Also time should perhaps be first parameter, while the rest of the parmeters have default values specified? (Similar how `toString` works for floats, while allowing precision and other properties to be specified).

I'd also like to make changes to how the log catogories are handled. Having to define the same category in two places (`BS_LOG_CATEGORY` and `BS_LOG_REGISTER_CATEGORY_NAME`) is not ideal. I was hoping to eventually handle this all through `BS_LOG_CATEGORY` - it would probably involve statics and have them auto-register the category when they're initialized (similar to `RTTIType` which uses `InitRTTIOnStart` helper static to register itself).

Also I'd prefer we get rid of the `LogCategoryName` class, and move that functionality into `Log` itself. Seems a bit too specific to be broken off into its own class.

guestnone:
I've merged the `LogCategoryName` class into `Log` and moved `bs_convert_timet_to_string` to be a part of `toString` series of functions (with proposed changes).

Regarding the `BS_LOG_REGISTER_CATEGORY_NAME`, I've originally though about doing this as part of the `BS_LOG_CATEGORY` but I didn't wanted to pollute the Log category structs and personally, I didn't have any idea how to do it reasonably during initialization.

BearishSun:
Thanks for making the changes, I'll see if I can whip something up regarding auto-registration.

BearishSun:
Merged. Thanks again. Ultimately I didn't find a good way to auto-register the categories, but I changed the approach a bit anyway.

PR #359, opened 2019-07-07
imgui plugin cmake

pgruenbacher:
I went ahead and make Cmake/Plugins directory to distinguish plugin cmake projects away from the module cmake files. there's no way to test the imgui plugin ("yet") but this should allow for testing the plugin build option.


pgruenbacher:
closing and re-opening to re-trigger travis

pgruenbacher:
ah good they pass. the macos fix worked.

pgruenbacher:
updated the plugin repo so that it doesn't use material.

Source/CMakeLists.txt
```diff
@@ -15,6 +15,9 @@ set_property(CACHE AUDIO_MODULE PROPERTY STRINGS OpenAudio FMOD Null)
 set(PHYSICS_MODULE "PhysX" CACHE STRING "Physics backend to use.")
 set_property(CACHE PHYSICS_MODULE PROPERTY STRINGS "PhysX" "Null")
 
+# include imgui plugin will be default false.
+set(INCLUDE_IMGUI_PLUGIN OFF CACHE BOOL "Include BsfImgui submodule plugin")
```
BearishSun: Could we rename the property to something like `BSF_PLUGIN_IMGUI`. I'm hoping to add `BSF` prefix to all the cmake properties eventually, so perhaps best to start now. And and we move the `PLUGIN` naming forward, so when sorted by name, we have all the plugin related stuff grouped together, when printed by cmake (or in CMake GUI). Since there are no categories it will help distinguish various cmake properties as their number grows.

Source/CMake/Plugins/BsImgui.cmake
```diff
@@ -0,0 +1,45 @@
+include(FetchContent)
+
+
```
BearishSun: Could we name the file `bsfImgui.cmake`, just to be consistent with the plugin name.

Source/CMake/Plugins/BsImgui.cmake
```diff
@@ -0,0 +1,45 @@
...
+# if INCLUDE_IMGUI_PLUGIN is true, then git clone the imgui plugin (with
+# recursive submodules). it will be cloned into the build directory where it
+# will be configured as bsfImgui target. other targets can then set bsfImgui
+# as their dependency in their respective cmake files.
+
+set(FETCHCONTENT_QUIET CACHE INTERNAL 0)
+
+FetchContent_Declare(
+  imgui
```
BearishSun: Where does this fetch to (i.e. what folder)?

Source/CMakeLists.txt
```diff
@@ -201,6 +204,14 @@ else() # Otherwise include only chosen ones
 	endif()
 endif()
 
+if (INCLUDE_IMGUI_PLUGIN)
+	# if true, then git clone the imgui plugin (with recursive submodules). it
```
BearishSun: Can probably get rid of this comment, as we have it in the file itself

BearishSun:
Thank you, looks good, I left some minor notes.

BearishSun:
(Also, not sure what's the best way to do a review on your imgui plugin repository. Not letting me comment on individual lines)

Source/CMakeLists.txt
pgruenbacher: done.

Source/CMake/Plugins/BsImgui.cmake
pgruenbacher: done.

Source/CMakeLists.txt
pgruenbacher: done

Source/CMake/Plugins/BsImgui.cmake
pgruenbacher: it will fetch to the Build/_deps/imgui-src folder and is built in the Build/_deps/imgui-build folder.

pgruenbacher:
i'll push updates in a sec. just make an issue with links to the lines you're concerned about.

BearishSun:
Thanks for making the changes, merged.

PR #361, opened 2019-07-08
Googletests

pgruenbacher:
notes
* fixed octree and quadtree memory issues
* fixed bitset static constexpr compilation issues (unless upgraded to c++17)
editor styling and linting notes to help me match code style

Reasons for migrating to googletests:
more featureful assertion tests and printing. e.g. for approximation tests, if the test fails, I want to see the actual difference in the values.
googletest and googlebenchmark are quite common in most c++ projects I find.
tests are auto-discovered. no need for having to register test functions.

caveats:
I couldn't get ctest working out of the gate with it, simply cause ctest tries to run each test on its own thread which is problematic for running tests on the same application thread and window.
still globbing all *_test.cpp files which allows for flexible discover of unit-tests without having them to all be in the same space. I imagine Marko doesn't like glob though. It works well for finding the test files though. and i like having test files next to their relevant source code instead of separate directory.

the PR is missing my ECS rendering and imgui tests. I still need to test how unit-tests would work with optional plugins. do plugins declare their unit-tests, or does the unittest.cmake simply have logic to handle the plugin code? I'd prefer to be able to run the unit test command once, not multiiple unit-test commands for each plugin as well. I do have it working with bsfRenderBeast plugin by explicity handling it in unittest.cmake, maybe I'll include imgui and ecs plugin tests as well once the imgui PR is approved.


pgruenbacher:
also threw in a quick google benchmarks (for cpu benchmarking, certainly not scene benchmarking which is a different benching scenario). More meant for benchmarking acceleration structures, static vectors, culling, etc.

pgruenbacher:
ok so it's in a ready state i guess, you could merge and squash if so desired. I treid getting headless travis tests working but I think it goes against the main duty of the travis being to build and deploy the assets. plus the opengl, x11 aps were failing on travis.

Source/Foundation/bsfUtility/Utility/BsQuadtree.h
```diff
@@ -420,8 +422,7 @@ namespace bs
 
 		private:
 			HNode mCurrentNode;
-			StaticAlloc<Options::MaxDepth * 4 * sizeof(HNode), FreeAlloc> mStackAlloc;
-			StaticVector<HNode, Options::MaxDepth * 4> mNodeStack;
+			std::vector<HNode> mNodeStack;
```
BearishSun: We don't use `std::vector`, but `Vector`. Also I avoid dynamic allocations as much as possible because 1. memory fragmentation, 2. Windows debug heap slows the debug mode too much with many of them (and it's useful to have enabled).

Not sure why StaticVector wasn't working, but SmallVector is it's newer and simpler replacement, so I'd prefer that if a change is necessary.

Source/Foundation/bsfUtility/Utility/BsBitstream.h
```diff
@@ -9,8 +9,17 @@
 #include "Math/BsVector3.h"
 #include "Utility/BsBitwise.h"
 
+
 namespace bs 
 {
+	using QuantType = uint8_t;
+	// see https://stackoverflow.com/questions/52355223/why-cant-a-static-constexpr-member-variable-be-passed-to-a-function
+	// static constexpr variables before c++17 can't be defined within the class.
+	static constexpr uint32_t BYTES_PER_QUANT = sizeof(QuantType);
```
BearishSun: They can be part of the class, they just need to be declared in the .cpp file as well, for certain cases (very compiler dependent so far). I'd prefer that option than moving it out of class.

Source/Foundation/bsfUtility/Testing/BsTestCameraFlyer.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include "BsPrerequisites.h"
+#include "Scene/BsComponent.h"
+#include "Math/BsMath.h"
+#include "Input/BsVirtualInput.h"
+
+namespace bs
+{
+	/** Component that controls movement and rotation of the scene object it's attached to. */
+	class CameraFlyer : public Component
```
BearishSun: Don't like having this here. It's 1. duplicated and 2. it's an example-only class (for now, until we add a higher level gameplay layer).

Also not really sure about adding these kind of complex tests here (esp. not in the utility layer, which is too low level for this functionality). Adding them to examples might make more sense, but even that I'm not sure about. I haven't thought enough about these types of unit tests yet, I'd rather handle them in a non-unit tests (e.g. an example project) for now.

Source/Foundation/bsfUtility/Private/UnitTests/BsUtility_bench.cpp
```diff
@@ -0,0 +1,69 @@
+//************************************ bs::framework - Copyright 2018 Marko Pintera **************************************//
```
BearishSun: I like the benchmarks, would be really nice to have these going for more systems. File name should probably be `BsUtilityBench` - similar naming concerns for `_test` files.

Source/Foundation/bsfUtility/Private/UnitTests/BsFileSystem_test.cpp
```diff
@@ -0,0 +1,274 @@
+//************************************ bs::framework - Copyright 2018 Marko Pintera **************************************//
+//*********** Licensed under the MIT license. See LICENSE.md for full terms. This notice is not to be removed. ***********//
+
+#include "Testing/BsTestSuite.h"
```
BearishSun: Would be nice it we didn't lose git history for these files, and could just see the changes done to accommodate google test.

Source/Foundation/bsfUtility/CMakeSources.cmake
```diff
@@ -180,18 +180,6 @@ set(BS_UTILITY_SRC_MATH
 	"bsfUtility/Math/BsLine2.cpp"
 )
 
-set(BS_UTILITY_INC_TESTING
-	"bsfUtility/Testing/BsTestSuite.h"
-	"bsfUtility/Testing/BsTestOutput.h"
-	"bsfUtility/Testing/BsConsoleTestOutput.h"
-)
-
-set(BS_UTILITY_SRC_TESTING
```
BearishSun: Filters for old files were removed, but haven't been added for new ones. VS or XCode won't display those files in their workflow without them.

CPPLINT.cfg
```diff
@@ -0,0 +1,3 @@
+# tabs should be tab, not space in bsf.
+# braces should be preceded by a newline
+filter=-whitespace/tab,-whitespace/braces
```
BearishSun: Is clang-format a perhaps a better option? Haven't really used cpplint, but some IDE's I work with already integrate clang-format.

.travis.yml
```diff
@@ -76,6 +76,11 @@ install:
 
   - cd ${TRAVIS_BUILD_DIR}
 
+before_script:
```
BearishSun: What do these do?

Source/Foundation/bsfUtility/Utility/BsBitstream.h
pgruenbacher: yea there just wasn't an associated cpp file so I wasn't sure if it was worth the effort

Source/Foundation/bsfUtility/Private/UnitTests/BsFileSystem_test.cpp
pgruenbacher: eh the only changes were changing BS_ASSERT to ASSERT_TRUE or ASSERT_EQ.

.travis.yml
pgruenbacher: oh i need to get rid of that. old legacy travis changes to try to get headless rendering working.

Source/Foundation/bsfUtility/Utility/BsQuadtree.h
pgruenbacher: ok i'll update.

BearishSun:
I looked over the code and made some comments.

Note that in general it's looking unlikely I'd be happy with the move to google test. Eventually I'd like to extend our unit testing system so it can work with higher-level unit tests, including tests for the renderer and multi-frame tests for gameplay features. Later perhaps export the testing framework to C# and integrate it into the editor.

All that work is probably going to happen sooner or later, so I'd be happier keeping the current test system and enhancing it with missing features.

BearishSun:
Oh yeah, I almost forgot. I really like the idea of adding benchmarks like you added. But again, we'll eventually need benchmarks for rendering (taking place over multiple frames) and we'll probably want to handle this using a custom system as well.

pgruenbacher:
ok I'll drop it. I def need the unit-tests working for development so I'll see if I can just get the existing unit-test code to work. maybe i'll look at how google code works.

pgruenbacher:
maybe there's some smaller MIT library examples that I can just copy-paste into the bsf code.

PR #363, opened 2019-07-09
add name to animation clip

pgruenbacher:
animation clips should have name for obvious reasons in terms of tracking.


Source/Foundation/bsfCore/Animation/BsAnimationClip.h
```diff
@@ -169,6 +169,15 @@ namespace bs
 		BS_SCRIPT_EXPORT(n:Events,pr:setter)
 		void setEvents(const Vector<AnimationEvent>& events) { mEvents = events; }
 
+
+		/** @copydoc setName() */
```
BearishSun: Base class `Resource` should have name getter/setters already.

pgruenbacher:
ah right you are, i'll make a new PR that just does the fbx importer set name

PR #365, opened 2019-07-10
set name of animation clip resource when importing

pgruenbacher:


BearishSun:
Good to have, thanks.

PR #367, opened 2019-07-12
Make CMake install the needed bslinc raw files

hhyyrylainen:
As discussed on discord, the raw bslinc assets are needed for compiling shaders, and the asset versions of shader includes are unnecessary (they don't work), this changes the cmake install to include the raw shader includes and exclude the compiled shader includes.

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -282,6 +282,12 @@ install(
 	DIRECTORY ../Data
 	DESTINATION ${DATA_DST_DIR}
 	PATTERN "../Data/Raw" EXCLUDE
+	PATTERN "../Data/Shaders/Includes" EXCLUDE
 	PATTERN ".version" EXCLUDE
 	PATTERN ".reqversion" EXCLUDE
+)
+
+install(
+    DIRECTORY ../Data/Raw/Shaders/Includes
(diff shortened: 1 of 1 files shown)


Source/CMakeLists.txt
```diff
@@ -282,6 +282,12 @@ install(
 	DIRECTORY ../Data
 	DESTINATION ${DATA_DST_DIR}
 	PATTERN "../Data/Raw" EXCLUDE
+	PATTERN "../Data/Shaders/Includes" EXCLUDE
 	PATTERN ".version" EXCLUDE
 	PATTERN ".reqversion" EXCLUDE
 )
+
+install(
+    DIRECTORY ../Data/Raw/Shaders/Includes
```
BearishSun: Make sure to use the same white space as the rest of the file (tabs in this case).

BearishSun:
Thanks for fixing this. Merged.

Source/CMakeLists.txt
hhyyrylainen: Oh, right. I wasn't expecting there to be tabs in a CMake file... my editor uses two spaces to indent cmake files so I made sure to add extra spaces, but completely missed the tabs

PR #368, opened 2019-07-12
Fix whitespace and newlines

pgruenbacher:
updated .editorconfig so this issues is fixed in the future.
otherwise just updates all files to not have trailing whitespace or missing newlines at end of file.
https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file


BearishSun:
Thanks, it's nice to have this standardized. 

I'm seeing additions of 1 byte to binary files, such as .png, .ttf and others. Can you avoid updating those?

pgruenbacher:
oops yea you're right.

pgruenbacher:
ok if you look at the file filter now, you should see that only text file extensions have been updated now.

BearishSun:
Thanks, merged.

PR #369, opened 2019-07-13
Fix BS_LOG and array related RTTI makros not working outside the bs-namespace

ataulien:
While integrating the new `BS_LOG`-makro into our codebase, I noticed it was directly accessing types from within the `bs` namespace which doesn't work when used from outside the `bs`-namespace. That was also a problem with the array related RTTI-macros, but we worked around the issue before.

This PR fixes both of those cases by explicitly using the types from the `bs`-namespace. The `BS_LOG`-makro was also wrapped in `do { ... } while(0)` to make the compiler complain about missing semicolons, of which there were quite a few.

### DIFF
--- a/Source/Foundation/bsfCore/Image/BsPixelUtil.cpp
+++ b/Source/Foundation/bsfCore/Image/BsPixelUtil.cpp
@@ -2515,13 +2515,13 @@ namespace bs
 	{
 		if (!isCompressed(options.format))
 		{
-			BS_LOG(Error, PixelUtility, "Compression failed. Destination format is not a valid compressed format.")
+			BS_LOG(Error, PixelUtility, "Compression failed. Destination format is not a valid compressed format.");
 			return;
 		}
 
 		if (src.getDepth() != 1)
 		{
-			BS_LOG(Error, PixelUtility, "Compression failed. 3D texture compression not supported.")
+			BS_LOG(Error, PixelUtility, "Compression failed. 3D texture compression not supported.");
 			return;
 		}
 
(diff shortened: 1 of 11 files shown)


BearishSun:
Thanks, merged. Just out of curiosity, which compiler complained about the semicolon?

ataulien:
Uh, I think you were a bit fast to merge that as the CI found some places not on my current platform which were still missing semicolons.

No compiler complained about them missing, but you can run into trouble with using makros without it inside if-statements without braces. More info on that here: http://www.bruceblinn.com/linuxinfo/DoWhile.html

BearishSun:
There were already braces.

I'm aware about potential cross-platform issues, but I prefer to let CI let me know rather than spend 30 mins compiling on all the platforms when there might be no issues at all.

ataulien:
The CI actually found some issues and I fixed them. Unfortunately you already clicked the merge button 😕 

Let's look at an example for when braces are not enough:

```cpp
#define FOO(x) foo(x); bar(x)

if (condition)
    FOO(x);
else // syntax error here
    ...;
```

Even using braces doesn't help:

```cpp
#define FOO(x) { foo(x); bar(x); }
```

Using this in an if statement would require that you omit the semicolon, which is counterintuitive:

```cpp
if (condition)
    FOO(x)
else
    ...
```

If you define FOO like this:

```cpp
#define FOO(x) do { foo(x); bar(x); } while (0)
```

then the following is syntactically correct:

```cpp
if (condition)
    FOO(x);
else
    ....
``` 

(Source: https://stackoverflow.com/questions/257418/do-while-0-what-is-it-good-for)

ataulien:
I opened #370 to take care of that.

PR #372, opened 2019-07-14
Make BS_LOG_VERBOSITY configurable from outside

ataulien:
This PR allows to set `BS_LOG_VERBOSITY` explicitly from outside, like via the build system, which is useful if you'd want to use `BS_LOG` for your own applications logging but don't need the logs internal to bsf.

To enable logging for a target in CMake, one can use:
```cmake
target_compile_definitions(MyApplication PRIVATE -DBS_LOG_VERBOSITY=LogVerbosity::Log)
```

It still defaults to the old behavior if one does not set it.


BearishSun:
Makes sense, thanks.

PR #373, opened 2019-07-17
Add CMake option to enable searching for DynLibs in an extra folder

hhyyrylainen:
This adds an CMake option that when specifies enables the DynLib
loader to look for the library in an additional folder.

I was unable to fully test this as it seems that even though I have my
libs in a sub folder, it seems that just loading them by name
works. Maybe my RPATH setting comes into play here? This is at least
so on Linux.


BearishSun:
Hi. Can you provide a use-case for this functionality?

hhyyrylainen:
Not having to clutter the same folder as the executable with a bunch of libraries is nice. I coded this before I found out that at least on my computer the ld open lib function seems to take RPATH into account.

BearishSun:
Makes sense. Although I'd prefer this path is something we provide to `DynLib` from higher level systems when plugins are being loaded, rather than adding it under the hood. Not sure if it's required to be a build option? I'd prefer we add a path to `START_UP_DESC` which can then used in `DynLib` calls for plugins, as required.

hhyyrylainen:
I suppose it makes more sense to have this in the START_UP_DESC. I was in the mindset of adding cmake flags, as I'm going to next do a PR for adding flags for enabling RTTI and exceptions, and leaving debug info alone. So I just made this a flag as well.

PR #375, opened 2019-07-17
Added CMake options to build with RTTI and exceptions, and skip debug info handling

hhyyrylainen:
I need these options in order to link BSF with my game. And I need the
files to be unstripped for my debug info processing scripts to work
properly.

Default values should have the same behaviour as before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,3 +23,8 @@ set(SCRIPT_API "None" CACHE STRING "Which scripting API/language to use, if any.
 set_property(CACHE SCRIPT_API PROPERTY STRINGS "None" "C#")
 
 set(SCRIPT_BINDING_GENERATION OFF CACHE BOOL "If true, script binding generation will be supported through a specialized build target. Enable this if you plan on modifying the scripting API. Requires the SBGen tool dependency. Only relevant if you have selected a SCRIPT_API other than \"None\".")
+
+set(PROCESS_DEBUG_INFO ON CACHE BOOL "If true debug info will be processed into separate file in release builds")
+
+set(ENABLE_EXCEPTIONS OFF CACHE BOOL "If true exceptions will be enabled when compiling")
+set(ENABLE_RTTI OFF CACHE BOOL "If true RTTI will be enabled when compiling")
--- a/Source/CMake/HelperMethods.cmake
+++ b/Source/CMake/HelperMethods.cmake
@@ -372,7 +372,7 @@ function(check_and_update_binary_deps DEP_PREFIX DEP_NAME DEP_FOLDER DEP_VERSION
 endfunction()
 
 function(strip_symbols targetName outputFilename)
-	if(UNIX)
+	if(UNIX AND PROCESS_DEBUG_INFO)
 		if(CMAKE_BUILD_TYPE STREQUAL Release)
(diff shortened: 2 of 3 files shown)


BearishSun:
Thanks, merged. I changed the parameter names a bit as I have a new naming scheme planned for cmake properties, to be applied retroactively at one point.

Also I don't think this enables exceptions on MSVC.

BearishSun:
Your username appears to break GitHub. Error message was shown on `bsf` front page `Unable to load latest commit`, when it had your name in it. I did a local rebase and force push of your name without the non-ascii characters, which appears to have fixed it.

hhyyrylainen:
I looked for the MSVC exception setting (/EH, based on this: https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=vs-2019) but didn't find it in the same file. But it does seem that that property is set:
```bash
$ ag /EH
Source/CMake/Properties.cmake
31:## Remove /EHsc from CMAKE_CXX_FLAGS for MSVC to disable exceptions
33:	if(CMAKE_CXX_FLAGS MATCHES "/EHsc")
34:		string(REPLACE "/EHsc" "/EHs-c-" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
```
I'm not on Windows so I can't test ATM. But something like this should take care of that:
```patch
diff --git a/Source/CMake/Properties.cmake b/Source/CMake/Properties.cmake
index f1997cc6f..5c7a582f4 100644
--- a/Source/CMake/Properties.cmake
+++ b/Source/CMake/Properties.cmake
@@ -29,7 +29,7 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
 
 ## Remove /EHsc from CMAKE_CXX_FLAGS for MSVC to disable exceptions
-if (MSVC)
+if (MSVC AND NOT BSF_ENABLE_EXCEPTIONS)
 	if(CMAKE_CXX_FLAGS MATCHES "/EHsc")
 		string(REPLACE "/EHsc" "/EHs-c-" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 	endif()
```

Re: my name breaking stuff on Github. I've not had that before (in fact this repo: https://github.com/Revolutionary-Games/Thrive has me as the latest committer and there doesn't seem to be any issues)

BearishSun:
I believe your name is not breaking stuff in that other repo as it's showing your username, which doesn't have non-ascii characters. Here it was showing your first and last name, which does. I'm guessing this difference is because I did a rebase, so it lists us both us authors and then fails to display your username and displays the name instead. Anyway, it would be appreciated if you could change your git config not to have that character in the name, as I usually rebase PRs.

hhyyrylainen:
You mistyped my email address, so maybe that's why it is showing the full name? I also can't see the commit on my profile, so Github thinks it isn't mine (https://github.com/GameFoundry/bsf/commits?author=hhyyrylainen&since=2019-06-30&until=2019-07-19).
I work on a bunch of different repos so I can't permanently edit my git config like that, though I can set a repo specific committer name if it is a problem (when the email is not typoed).

Can you check that .patch I posted or should I make another PR?

BearishSun:
Perhaps it was some other issue, I guess we'll see next time. I'll add your change for MSVC, thanks.

PR #384, opened 2019-08-04
Feature: extended `GpuResourceData` interface with `setData` function

nemerle:
Refactor: introduce `verifyLockAndThread` helper function to reduce repetition.

This should make modifying contents easier and reduce the need to allocate
fully blown PixelData to implement simple operations ( flipping images etc. )

And yes I'm aware of `PixelUtil::mirror`, the point of SetData is mostly in inter-operability with external sources of image data :smile:


BearishSun:
Looks good, useful to have.

PR #385, opened 2019-08-04
Feature: Introduce `writeSubData` in `Texture` API

nemerle:
This is mostly a proposal and a call for help from D3D11 and Vulkan gurus out there.

What is missing:
- Vulkan support
- D3D11 support
- Optional written-area-within-target checks.


BearishSun:
I'd prefer the new method is also named `writeData` and is just an overload. Other than that I think it's fine, as long as the underlying API's all support this.

nemerle:
> I'd prefer the new method is also named `writeData` and is just an overload. Other than that I think it's fine, as long as the underlying API's all support this.

I've originally named those methods `writeData`, but was unsure how well the `@copydoc` worked with overridden functions.

PR #388, opened 2019-08-08
Add color property to renderables

mworchel:
This PR introduces a `color` property to renderables, more specifically it adds a multiplicative base color which is set on a per-instance basis. A typical use case is having many object instances sharing the same mesh and basically the same material, except for slight color variations. With the classical pipeline you would need to create one material per instance, resulting in one instanciation of the shader per instance (which is slow when having a large amount of objects). 

Although this change arguably serves a niche use case, I find adding this property benefits the framework and integrates really well with the rest of the per object properties available in the shaders.

The PR can be seen as a first basic implementation and currently only covers renderables, although it is probably also applicable to shaded particles and decals. I'm of course open to feedback, if anything needs to be change or extended.

### DIFF
--- a/Data/Raw/Shaders/Default.bsl
+++ b/Data/Raw/Shaders/Default.bsl
@@ -17,7 +17,7 @@ shader Surface
 			out float OutGBufferD : SV_Target4)
 		{
 			SurfaceData surfaceData;
-			surfaceData.albedo = float4(0.05f, 0.05f, 0.05f, 1.0f);
+			surfaceData.albedo = float4(0.05f, 0.05f, 0.05f, 1.0f) * gColor;
 			surfaceData.worldNormal.xyz = input.tangentToWorldZ;
 			surfaceData.roughness = 1.0f;
 			surfaceData.metalness = 0.0f;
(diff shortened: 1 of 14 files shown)


BearishSun:
Thanks for the PR. While I do understand your use-case I'm afraid I have some core concerns regarding the design:
 - The Renderable is meant to be decoupled from any particular Material implementation, and making the assumption the color tint might be needed isn't ideal.
 - The design intent of the Material class is to encapsulate any shader properties, and this breaks that intent. If the Material instances are too slow or too cumbersome to create we should instead work on improving those aspects of Material. In the current system Materials with different properties will still batch together, but that can be further improved with support for hardware instancing mentioned in the next point.
 - This doesn't solve the problem in the general case, as users might have need to make other properties instanced as well. We should work on a solution that extends Material with support for such properties, and handle any additional work under the hood. This was discussed recently on discord. On top of being able to work on generic material properties, this approach also does not require separate implementations for renderable, decals and particle systems.

mworchel:
I completely agree that a more generic design for passing per-instance parameters would be the better solution. I wasn't aware of the ongoing discussion in the discord about this topic but now I'm at least some kind of relieved that instancing is on the table. Especially with the design intents of the `Renderable` and `Material` I understand, why my design wouldn't really fit that well. I'm closing this PR now and maybe I can contribute something to the more general solution in the future.

PR #399, opened 2019-09-05
Added external window resize

### LINKED ISSUE #382: External windows need method for notifying about resize
With an external window handle created bs::RenderWindow, resizing doesn't work properly. There needs to be a new set of methods for notifying the window that "your size / fullscreen has changed". It is kinda possible on windows to call the `RenderWindow::resize` method to make it somewhat work (resizing will also resize in the wrong direction), but on windows that locks up the program entirely.

hhyyrylainen:
With these methods when using an external window handle, the window can be resized properly. Only tested on Linux as I couldn't get the master branch version running on my laptop: https://pastebin.com/eT9gZwv8

fixes #382

### DIFF
--- a/Source/Foundation/bsfCore/RenderAPI/BsRenderWindow.cpp
+++ b/Source/Foundation/bsfCore/RenderAPI/BsRenderWindow.cpp
@@ -364,6 +364,46 @@ namespace bs
 		}
 	}
 
+	void RenderWindow::_onExternalResize(UINT32 width, UINT32 height)
+	{
+		RenderWindowProperties& props = getMutableProperties();
+		props.width = width;
+		props.height = height;
+		_notifyWindowEvent(WindowEventType::Resized);
+	}
+
+	void RenderWindow::_onExternalMove(INT32 top, INT32 left)
+	{
+		RenderWindowProperties& props = getMutableProperties();
+		props.top = top;
+		props.left = left;
+		_notifyWindowEvent(WindowEventType::Moved);
+	}
+
+	void RenderWindow::_onExternalFocus(bool focused)
+	{
+		if(focused)
+		{
+			_notifyWindowEvent(WindowEventType::FocusReceived);
+		}
+		else
+		{
+			_notifyWindowEvent(WindowEventType::FocusLost);
+		}
+	}
+
+	void RenderWindow::_onExternalMaximized(bool maximized)
+	{
+		if(maximized)
+		{
+			_notifyWindowEvent(WindowEventType::Maximized);
+		}
+		else
+		{
+			_notifyWindowEvent(WindowEventType::Restored);
+		}
+	}
+
 	/************************************************************************/
 	/* 								SERIALIZATION                      		*/
 	/************************************************************************/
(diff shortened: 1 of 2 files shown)


hhyyrylainen:
I did a quick test on windows and resizing seemed to work there as well.

BearishSun:
Is the swap chain properly resized as well? Because `bs::RenderWindow::_notifyWindowEvent` isn't used on Windows or Linux normally. It's core-thread counterpart `bs::ct::RenderWindow::_notifyWindowEvent` is called instead (as the main window loop runs on the core thread on those OS's).

This is important because in both methods it calls `_windowMovedOrResized` virtual method on their respective `RenderWindow` types. And on Windows & Linux, `bs::RenderWindow::_windowMovedOrResized` is not implemented. It's only implemented on the core thread, where it handles swap chain resize (also perhaps some other stuff).

hhyyrylainen:
As far as I can tell the render target gets resized. As the render time goes from 10ms to 30ms when resizing to 4K resolution. Is there something extra that should be added to the notification function?

hhyyrylainen:
I tried adding `bs::gCoreThread().queueCommand([core{getCore()}](){ core->_windowMovedOrResized();});` to the notify method. But it doesn't seem to have any effect. Maybe marking the properties dirty already does the right thing?

BearishSun:
Perhaps. I'd put a breakpoint `_windowMovedOrResized` and see if and how it triggers (I'm guessing some OS event is still triggering it?).

Anyway, this is in any case better than no external resize handling so I'll merge. I'm just worried it's not working by accident.

BearishSun:
Note it can happen the render time goes up as all the render targets are proper size, but the final swap chain target might remain small. But in that case you'll see a visible pixelation in your app.

hhyyrylainen:
Thanks.

I put a breakpoint on _windowMovedOrResized and it is infact not triggered when I don't have that extra line. But seeing the implementation (at least on Linux) it doesn't seem to actually do much:
```c++
	void LinuxRenderWindow::_windowMovedOrResized()
	{
		if (!mWindow)
			return;

		RenderWindowProperties& props = mProperties;
		if (!props.isFullScreen) // Fullscreen is handled directly by this object
		{
			props.top = mWindow->getTop();
			props.left = mWindow->getLeft();
			props.width = mWindow->getWidth();
			props.height = mWindow->getHeight();
		}
	}
```
So I don't think it's super critical that that doesn't get called as the properties should be updated anyway. Right?

Anyway I can't really see any pixelation with the 4K screenshot compared to the 720p: https://imgur.com/a/1PV5zm4

BearishSun:
The method rebuilds the swap chain on DirectX and Vulkan, it might be broken on those.

hhyyrylainen:
Alright. I'll do a further PR if that is a problem and I find a fix.

BearishSun:
I think the best way to solve this issue is to move all these external methods to the core `RenderWindow`. Then call them through `bs::gCoreThread().queueCommand()`. Since on Windows & Linux all the window/OS operations are reported on the core thread already, it's more consistent with that behavior and less likely to cause issues.

In fact you might be able to call `_notifyWindowEvent` directly that way, without the additional methods. (Although there might be a delay for the width/height/etc. to sync from core -> sim in that case, if that's important).

hhyyrylainen:
You were correct, this does not work correctly on windows. I'll try your suggestion.

hhyyrylainen:
After a bit of testing I discovered that only directx has problems with my current method. OpenGL and Vulkan resize just fine. Changing my code to this:
```c++
    // Notify bsf
    bs::gCoreThread().queueCommand([coreWindow = BSFWindow->getCore()]() {
        coreWindow->_notifyWindowEvent(bs::WindowEventType::Resized);
    });

    BSFWindow->_onExternalResize(width, height);
```
makes directx resize work as well.

